Issue #256, opened 2014-11-28
Isolate token factory from concrete token generation


backbone87:
Currently the generic token factory is used in all token generation applications. Also if you want to setup a token factory (for example in a DI container) you need to know about the routes for the tokens, which may not be known in the defining code entity.

Also the token factory is used by some actions directly:
https://github.com/Payum/Payum/blob/master/src/Payum/Paypal/ExpressCheckout/Nvp/Action/FillOrderDetailsAction.php#L19
If you use the PaymentFactory of express checkout you cant use the notify feature, because of this.

Proposed changes:
- Rename GenericTokenFactoryInterface to TokenProviderInterface
- TokenProviderInterface should not extend TokenFactoryInterface
- concrete TokenProviderInterface impls should use a TokenFactoryInterface for creation of tokens
- Introduce new TokenProviderAware interface to be implemented in actions that need to generate tokens
- Introduce new TokenProviderExtension which injects a TokenProvider into TokenProviderAware actions

makasim:
TokenFactory now uses composition - with https://github.com/Payum/Payum/pull/311

makasim:
So only thing is left is a token provider extension

PR #267, opened 2014-11-30
WIP Token provider

backbone87:
impl of #256 

tests missing (i am not that good at test driven dev)


makasim:
close in favor https://github.com/Payum/Payum/pull/314

PR #288, opened 2014-12-29
Add redirecting animation

kayue:
Demo: http://jsbin.com/sobogodifa/1/


makasim:
@kayue do you have plans to finish it?

Issue #287, opened 2014-12-24
USERACTION=commit needs to be useraction=commit


makasim:
This does appear to be a bug in Payum - in the URL, the string USERACTION=commit needs to be useraction=commit. i.e. all in lower case. This has been confirmed by Paypal support.

Discussing started here: https://github.com/Payum/PayumBundle/issues/153

vkartaviy:
:+1:  Just stumbled upon this... :(

PR #296, opened 2015-01-03
Paypal request parameters should be lowercased

vkartaviy:
Fix for #287


makasim:
There is one test failed, Could you please fix it?

vkartaviy:
Yep, doing that right now :)

PR #303, opened 2015-01-13
Fix BOM at start of stream

winzou:
Blank utf8 character that makes json_decode fail because of syntax error.
cf. http://en.wikipedia.org/wiki/Byte_order_mark

### DIFF
--- a/src/Payum/Core/Bridge/Buzz/JsonResponse.php
+++ b/src/Payum/Core/Bridge/Buzz/JsonResponse.php
@@ -13,9 +13,16 @@ class JsonResponse extends Response
      */
     public function getContentJson()
     {
-        $json = json_decode($this->getContent());
+        $content = $this->getContent();
+
+        // Remove unexpected utf8 BOM
+        if(substr($content, 0, 3) == pack('CCC', 239, 187, 191)) {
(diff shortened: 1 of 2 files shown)


makasim:
is it possible to add a test for this case?

makasim:
for the reference https://github.com/guzzle/guzzle/pull/422. Btw where did you face it? what was the service?

winzou:
Added tests + an optional parameter to decode json in array instead of stdclass.

I'm working for Limonetik http://www.limonetik.com/en and they have a .NET API server which adds this BOM.

makasim:
so maybe it would be better to put all this stuff to your lib. Practice to add BOM considered as bad practice. And I am tend to agree with what was said here guzzle/guzzle#422. Even more I want to remove this custom response class as the main goal to migrate to PSR-7 client. 

We can merge that for now, but in future things may change and you will have to adjust your lib.

winzou:
Sure I can put it in my lib. I just thought it would have been useful for others, but if you move out of this class then it does not make sense.

makasim:
More... we can migrate to Guzzle and in that case you can reuse the plugin mentioned there that solve the problem

PR #307, opened 2015-01-28
Fix error in payment name

smarques:
In http://payum.org/doc/0.14/PaypalProCheckoutNvp/get-it-started the description links to the right paypal page, but the linked name is still paypal express, it should be pay pal pro checkout.


makasim:
Thanks

smarques:
:+1:

PR #308, opened 2015-01-31
[Authorize.NET] Update order fields name.

yannkechichian:
Hi guys,

Not sure whether Authorize.NET updated their SDK but this PR fixes some fields name in the order details as per this definition (line 80): https://github.com/AuthorizeNet/sdk-php/blob/master/lib/AuthorizeNetAIM.php

Currently, the invoice number, email and customer ID are not populated or causing an error if these fields are set as required.


makasim:
@yannkechichian thanks.

PR #313, opened 2015-02-11
Add Propel2 bridge

lovenunu:
This PR adds the support of Propel 2 as a storage.
The steps to follow to setup the Propel 2 storage are in the ```storage.md``` file.
The tests are based with Propel master, but as the user has to generate the base classes, the bridge will work with previous version of Propel2.

If you want to try it live, you can use this snippet: 
```php
<?php

include "vendor/autoload.php";

use Propel\Runtime\Propel;
use Propel\Runtime\Connection\ConnectionManagerSingle;
use Payum\Core\Bridge\Propel2\Storage\Propel2Storage;

$serviceContainer = Propel::getServiceContainer();
$serviceContainer->setAdapterClass('default', 'mysql');
$manager = new ConnectionManagerSingle();
$manager->setConfiguration(array (
    'dsn'      => 'mysql:host=localhost;dbname=payum',
    'user'     => 'root',
    'password' => 'root',
));
$serviceContainer->setConnectionManager('default', $manager);


$orderStorage = new Propel2Storage(
    "Payum\\Core\\Bridge\\Propel2\\Model\\Order"
);

$tokenStorage = new Propel2Storage(
    "Payum\\Core\\Bridge\\Propel2\\Model\\Token"
);

$order = $orderStorage->create();

if (!$order instanceof \Payum\Core\Bridge\Propel2\Model\Order) {
    throw new Exception;
}

$orderStorage->update($order);

if (1 <= \Payum\Core\Bridge\Propel2\Model\OrderQuery::create()->count()) {
    echo '\o/';
}
```


lovenunu:
Oh. Propel 2 needs php 5.4 to run, so I think it's better to remove it from composer dependencies.

makasim:
@lovenunu why? Just add skipp tests if php <= 5.4

makasim:
I am a bit confused. If I want to add a custom method to `Payum\Core\Bridge\Propel2\Model\PayumOrder`. I have to extend this model or modify this class (it sounds really weird )? 

Do propel generates classes to  `src/Payum/Core/Bridge/Propel2/Model/Base` folder? Looks strange...

lovenunu:
It isn't a tests problem, on php 5.3, composer can't resolve dependencies as propel/propel require php >= 5.4

Sorry for not being clear. Propel generates the classes in a ```Base``` folder, the one that are in the PR are those where you can add custom methods.

src/Payum/Core/Bridge/Propel2/Model/.gitignore
```diff
@@ -0,0 +1,2 @@
+Base/
+Map/
```
makasim: the .gitignore should be in root folder, shoudn't it? If so the root folder is `src/Payum/Core/` in case you installed  `payum/core` and `/` in case of `payum/payum`. Should be added to both files.

src/Payum/Core/Bridge/Propel2/Model/PayumOrder.php
```diff
@@ -0,0 +1,20 @@
...
+ * Skeleton subclass for representing a row from the 'payum_order' table.
+ *
+ *
+ *
+ * You should add additional methods to this class to meet the
+ * application requirements.  This class will only be generated as
+ * long as it does not already exist in the output directory.
+ *
+ */
+class PayumOrder extends BasePayumOrder
```
makasim: There is not need to prefix it with `Payum`. Name it just `Order`

src/Payum/Core/Bridge/Propel2/Resources/config/propel.xml
```diff
@@ -0,0 +1,26 @@
...
+        <runtime>
+            <defaultConnection>default</defaultConnection>
+            <connection>default</connection>
+        </runtime>
+        <generator>
+            <defaultConnection>default</defaultConnection>
+            <connection>default</connection>
+        </generator>
+    </propel>
+</config>
```
makasim: is it needed for tests or could be reused somehow in apps?

src/Payum/Core/Bridge/Propel2/Resources/config/schema.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Based on Propel1 Bridge schema -->
+<database defaultIdMethod="native" name="default">
+    <vendor type="mysql">
+        <parameter name="Engine" value="InnoDB"/>
+        <parameter name="Charset" value="utf8"/>
+    </vendor>
+
+    <table name="payum_order" namespace="Payum\Core\Bridge\Propel2\Model">
```
makasim: The doc in FOSUserBundle says next:

```
If you don't want to add your own logic in your user class, you can simply use FOS\UserBundle\Propel\User as user class and you don't have to create another class.

If you want to add your own fields, you can extend the model class by overriding the database schema. Just copy the Resources/config/propel/schema.xml file to app/Resources/FOSUserBundle/config/propel/schema.xml, and customize it to fit your needs.
```

Is it applies for these models too?

src/Payum/Core/composer.json
```diff
@@ -68,6 +68,7 @@
         "doctrine/mongodb-odm": "If you want to store models to mongo doctrin2 ODM",
         "zendframework/zend-db": "If you want to store models to Zend Db ORM",
         "propel/propel1": "If you want to store models to Propel1 ORM",
+        "propel/propel": "If you want to store models to Propel2 ORM",
```
makasim: same has to be add to composer.json in root folder.

src/Payum/Core/Bridge/Propel2/Resources/config/schema.xml
makasim: [They also](https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Resources/config/propel/schema.xml#L2) defines a namespace on the database level one time. on the table level they just set a class name and table one

src/Payum/Core/Bridge/Propel2/Resources/install/default.sql
```diff
@@ -0,0 +1,44 @@
...
+CREATE TABLE `payum_token`
+(
+    `id` INTEGER NOT NULL AUTO_INCREMENT,
+    `hash` VARCHAR(255),
+    `details` MEDIUMBLOB,
+    `after_url` VARCHAR(255),
+    `target_url` VARCHAR(255),
+    `payment_name` VARCHAR(255),
+    PRIMARY KEY (`id`)
+) ENGINE=InnoDB CHARACTER SET='utf8';
```
makasim: I'd also prefer to do another way. Build a database by schema. it should be possible. and drop this file.

src/Payum/Core/Bridge/Propel2/Resources/config/propel.xml
lovenunu: No, this is just needed for propel's ```build``` and ```sql:build``` commands.

makasim: Could we ask a developer to create such file in his code, and generate a database using schema? Looks better to me...

src/Payum/Core/composer.json
lovenunu: It already is in the root composer.json :)

src/Payum/Core/Bridge/Propel2/Resources/config/propel.xml
lovenunu: Then should we forget about having the model in the core ?

src/Payum/Core/Bridge/Propel2/Model/PayumOrder.php
lovenunu: No problem, but what about the sql table name ? Should I keep the payum_ prefix for ```payum_(order|token)``` to allow developpers to have a 1 database app or not ?
Here's a real example: In Thelia we already have an ```order``` table, and many of our users are on shared hostings, and only have 1 database, so we need to have a different name to allow our users to use payum with the propel 2 bridge.

src/Payum/Core/Bridge/Propel2/Model/.gitignore
lovenunu: Excuse me, I just did the same as Propel 1 bridge :)

src/Payum/Core/Bridge/Propel2/Resources/config/propel.xml
makasim: @lovenunu sorry I dont understand

src/Payum/Core/Bridge/Propel2/Resources/config/propel.xml.dist
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="ISO-8859-1" standalone="no"?>
```
lovenunu: I kept the propel.xml file for easy distribution.

makasim: Is it really worth it? You can find this file in the propel doc http://propelorm.org/documentation/02-buildtime.html or
here http://propelorm.org/documentation/10-configuration.html

What if they change something in schema we have to do same in this file and this the main reason not keep it here.

src/Payum/Core/composer.json
makasim: I see, sorry

.gitignore
```diff
@@ -2,4 +2,6 @@ composer.phar
 composer.lock
 cov
 vendor
-bin
\ No newline at end of file
+bin
+src/Payum/Core/Bridge/Propel2/Model/Base
+src/Payum/Core/Bridge/Propel2/Model/Map
```
makasim: You have to also add this to core folder (the file .gitignore must be created)

lovenunu: It already is in the commit.

makasim:
@lovenunu thanks!

lovenunu:
@makasim You're welcome ! :)

Issue #320, opened 2015-02-27
Cancel recurring payments


alexandernst:
I'm wondering if a recurring payment can be canceled from Payum, or that is something that can be done only from PayPal's profile panel. Can it be done?

makasim:
not yet, It could be done this way now https://github.com/makasim/PayumBundleSandbox/blob/master/src/Acme/PaypalExpressCheckoutBundle/Controller/RecurringPaymentExamplesController.php#L177

but would be good to wrap to cancel action so it looks like: $payment->execute(new Cancel($model));

alexandernst:
Let's continue this in the PR :)

PR #321, opened 2015-02-28
Add skeleton for canceling recurring payments

alexandernst:
This is a WIP for #320

### DIFF
--- a/src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/CancelRecurringPaymentsProfileAction.php
+++ b/src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/CancelRecurringPaymentsProfileAction.php
@@ -1,0 +1,42 @@
+<?php
+namespace Payum\Paypal\ExpressCheckout\Nvp\Action\Api;
+
+use Payum\Core\Request\Cancel;
+use Payum\Paypal\ExpressCheckout\Nvp\Api;
+use Payum\Core\Bridge\Spl\ArrayObject;
+use Payum\Core\Exception\RequestNotSupportedException;
+
+class CancelRecurringPaymentsProfileAction extends BaseApiAwareAction
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function execute($request)
+    {
+        /** @var $request Cancel */
+        RequestNotSupportedException::assertSupports($this, $request);
+
+        $model = ArrayObject::ensureArrayObject($request->getModel());
+        $model->validateNotEmpty(array('PROFILEID', 'ACTION', 'BILLINGPERIOD'));
+
+        $model['ACTION'] = Api::RECURRINGPAYMENTACTION_CANCEL;
+
+        $model->replace(
+            $this->api->manageRecurringPaymentsProfileStatus((array) $model)
+        );
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function supports($request)
+    {
+        if (false == ($request instanceof Cancel && $request->getModel() instanceof \ArrayAccess)) {
+            return false;
+        }
+
+        // it must take into account only recurring payments, common payments must be cancelled by another action.
+        $model = $request->getModel();
+        return isset($model['BILLINGPERIOD']);
+    }
+}
--- a/src/Payum/Paypal/ExpressCheckout/Nvp/Tests/Action/Api/CancelRecurringPaymentsProfileActionTest.php
+++ b/src/Payum/Paypal/ExpressCheckout/Nvp/Tests/Action/Api/CancelRecurringPaymentsProfileActionTest.php
@@ -1,0 +1,208 @@
+<?php
+namespace Payum\Paypal\ExpressCheckout\Nvp\Tests\Action\Api;
+
+use Payum\Core\Request\Cancel;
+use Payum\Paypal\ExpressCheckout\Nvp\Action\Api\CancelRecurringPaymentsProfileAction;
+use Payum\Paypal\ExpressCheckout\Nvp\Request\Api\ManageRecurringPaymentsProfileStatus;
+use Payum\Core\Tests\GenericActionTest;
+
+class CancelRecurringPaymentsProfileActionTest extends GenericActionTest
+{
+
+    /**
+     * @var Generic
+     */
+    protected $requestClass = 'Payum\Core\Request\Cancel';
+
+    /**
+     * @var ActionInterface
+     */
+    protected $actionClass = 'Payum\Paypal\ExpressCheckout\Nvp\Action\Api\CancelRecurringPaymentsProfileAction';
+
+    public function provideSupportedRequests()
+    {
+
+        return array(
+            array(new $this->requestClass(array('BILLINGPERIOD' => 'foo'))),
+            array(new $this->requestClass(new \ArrayObject(array('BILLINGPERIOD' => 'foo'))))
+        );
+    }
+    public function provideNotSupportedRequests()
+    {
+        return array(
+            array('foo'),
+            array(array('foo')),
+            array(new \stdClass()),
+            array(new $this->requestClass('foo')),
+            array(new $this->requestClass(new \stdClass())),
+            array($this->getMockForAbstractClass('Payum\Core\Request\Generic', array(array()))),
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function shouldBeSubClassOfBaseApiAwareAction()
+    {
+        $rc = new \ReflectionClass($this->actionClass);
+
+        $this->assertTrue($rc->isSubclassOf('Payum\Paypal\ExpressCheckout\Nvp\Action\Api\BaseApiAwareAction'));
+    }
+
+    /**
+     * @test
+     */
+    public function couldBeConstructedWithoutAnyArguments()
+    {
+        new CancelRecurringPaymentsProfileAction();
+    }
+
+    /**
+     * @test
+     */
+    public function shouldSupportManageRecurringPaymentsProfileStatusRequestAndArrayAccessAsModel()
+    {
+        $action = new CancelRecurringPaymentsProfileAction();
+
+        $this->assertTrue(
+            $action->supports(new Cancel(array('BILLINGPERIOD' => 'foo')))
+        );
+
+        $this->assertTrue(
+            $action->supports(new Cancel(new \ArrayObject(array('BILLINGPERIOD' => 'foo'))))
+        );
+    }
+
+    /**
+     * @test
+     */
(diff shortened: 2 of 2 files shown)


alexandernst:
This could be the base/skeleton code for canceling recurring payments.

As you said, the actual code is here https://github.com/makasim/PayumBundleSandbox/blob/master/src/Acme/PaypalExpressCheckoutBundle/Controller/RecurringPaymentExamplesController.php#L177 and it just needs to be adapted.

Tests are missing too.

src/Payum/Paypal/ExpressCheckout/Nvp/Request/Api/CancelRecurringPaymentProfile.php
```diff
@@ -0,0 +1,8 @@
+<?php
+namespace Payum\Paypal\ExpressCheckout\Nvp\Request\Api;
+
+use Payum\Core\Request\Generic;
+
+class CancelRecurringPaymentProfile extends Generic
```
makasim: there is no need for the specific request, use Cancel one from core.

src/Payum/Paypal/ExpressCheckout/Nvp/Api.php
```diff
@@ -426,6 +426,24 @@ public function doExpressCheckoutPayment(array $fields)
      *
      * @return array
      */
+    public function cancelRecurringPaymentsProfile(array $fields)
+    {
+        $request = new FormRequest();
+        $request->setFields($fields);
+
+        $request->setField('METHOD', 'CancelRecurringPaymentsProfile');
```
makasim: cannot find this `CancelRecurringPaymentsProfile` Paypal method. Could you point me to the doc

src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/CancelRecurringPaymentProfileAction.php
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public function supports($request)
+    {
+        return
+            $request instanceof CancelRecurringPaymentProfile &&
+            $request->getModel() instanceof \ArrayAccess
+        ;
+    }
```
makasim: must be like this:

```php
<?php

if (false == ($request instanceof Cancel && $request->getModel() instanceof \ArrayAccess)) {
    return false;
}

return isset($model['BILLINGPERIOD']); // it must take into account only recurring payments, common payments must be cancelled by another action.

             &&
            
            
        ;

``

File: src/Payum/Paypal/ExpressCheckout/Nvp/Api.php (hunk 2 above)
Comment (alexandernst) [inline src/Payum/Paypal/ExpressCheckout/Nvp/Api.php]: Oops, indeed it doesn't exist. I misread the docs. This should be a call to `ManageRecurringPaymentsProfileStatus` with 2 arguments, `PROFILEID` and `ACTION`. Let me make a patch (probably this afternoon)

**Comment by makasim** [comment]
@alexandernst ping

**Comment by alexandernst** [comment]
Hi Maksim! Sorry for delaying this,  had to do a few other things. Let me
see if I can have anything by tomorrow.

**Comment by alexandernst** [comment]
@makasim Hi! I just pushed another commit. I think it's missing something, but I'm not really sure what is it.

File: src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/CancelRecurringPaymentProfileAction.php [hunk 4]
```diff
@@ -0,0 +1,40 @@
...
+     */
+    public function execute($request)
+    {
+        /** @var $request CancelRecurringPaymentProfile */
+        RequestNotSupportedException::assertSupports($this, $request);
+
+        $model = ArrayObject::ensureArrayObject($request->getModel());
+        $model['ACTION'] = Api::RECURRINGPAYMENTACTION_CANCEL;
+
+        $model->validateNotEmpty(array('PROFILEID', 'ACTION'));
```
makasim: do it before you set `$model['ACTION']`.

src/Payum/Paypal/ExpressCheckout/Nvp/Api.php
```diff
@@ -426,6 +426,24 @@ public function doExpressCheckoutPayment(array $fields)
      *
      * @return array
      */
+    public function cancelRecurringPaymentsProfile(array $fields)
```
makasim: there is no need for this API method. use `manageRecurringPaymentsProfileStatus` one

makasim:
Except two small comments everything looks good. Would you able to add some tests?

alexandernst:
Every time I do a commit, the code gets smaller and smaller! hehehe :)

Yeah sure, let me see how tests are done and I'll push another commit in a few hours

alexandernst:
Well... for some reason `CancelRecurringPaymentsProfileAction` thinks it doesn't support `Cancel` requests, but it's clear that at the moment I'm not using my brain, so I think the best thing would be to have some rest and continue tomorrow @makasim 

Btw, any ideas that would help me?

src/Payum/Paypal/ExpressCheckout/Nvp/PaymentFactory.php
```diff
@@ -77,6 +77,7 @@ public function createConfig(array $config = array())
             'payum.action.api.get_express_checkout_details' => new GetExpressCheckoutDetailsAction(),
             'payum.action.api.get_transaction_details' => new GetTransactionDetailsAction(),
             'payum.action.api.do_express_checkout_payment' => new DoExpressCheckoutPaymentAction(),
+            'payum.action.api.cancel_recurring_payment_profile' => new ManageRecurringPaymentsProfileStatusAction(),
```
makasim: remove it

alexandernst:
@makasim Still failing :(
The problem is that `CancelRecurringPaymentsProfileAction` doesn't accept `Cancel` requests, as per the error message: 

    Payum\Paypal\ExpressCheckout\Nvp\Tests\Action\Api\CancelRecurringPaymentsProfileActionTest::shouldCallApiManageRecurringPaymentsProfileStatusMethodWithExpectedRequiredArguments
    Payum\Core\Exception\RequestNotSupportedException: Action CancelRecurringPaymentsProfileAction is not supported the request Cancel{model: ArrayObject}.

src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/CancelRecurringPaymentsProfileAction.php
```diff
@@ -0,0 +1,41 @@
...
+     * {@inheritDoc}
+     */
+    public function supports($request)
+    {
+        if (false == ($request instanceof Cancel && $model = $request->getModel() instanceof \ArrayAccess)) {
+            return false;
+        }
+
+        // it must take into account only recurring payments, common payments must be cancelled by another action.
+        return isset($model['BILLINGPERIOD']);
```
alexandernst: @makasim This is actually breaking tests like

        $action = new CancelRecurringPaymentsProfileAction();
        $this->assertTrue(
            $action->supports(new Cancel($this->getMock('ArrayAccess')))
        );

Maybe a better idea would be to check if `BILLINGPERIOD` is set in the `execute` method?

makasim: nope, because there maybe a cancel request for common payment (not recurring one) and in that case second action would name work.  It must be in supports method. You can use any other fields that allows you different common payment from recurring one.

alexandernst: The thing is that `getMock` won't play nice with the `isset($model['BILLINGPERIOD'])` condition in the `supports` method. Any ideas how to do it properly?

src/Payum/Paypal/ExpressCheckout/Nvp/Tests/Action/Api/CancelRecurringPaymentsProfileActionTest.php
```diff
@@ -0,0 +1,173 @@
+<?php
+namespace Payum\Paypal\ExpressCheckout\Nvp\Tests\Action\Api;
+
+use Payum\Core\Request\Cancel;
+use Payum\Paypal\ExpressCheckout\Nvp\Action\Api\CancelRecurringPaymentsProfileAction;
+use Payum\Paypal\ExpressCheckout\Nvp\Request\Api\ManageRecurringPaymentsProfileStatus;
+
+class CancelRecurringPaymentsProfileActionTest extends \PHPUnit_Framework_TestCase
```
makasim: Try to extend from https://github.com/Payum/Payum/blob/master/src/Payum/Core/Tests/GenericActionTest.php class. and configure providers methods.

src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/CancelRecurringPaymentsProfileAction.php
makasim: so the problem appear only in tests, is it? Lets continue discussion there: https://github.com/Payum/Payum/pull/321/files#r26213511

src/Payum/Paypal/ExpressCheckout/Nvp/Tests/Action/Api/CancelRecurringPaymentsProfileActionTest.php
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * @test
+     */
+    public function shouldSupportManageRecurringPaymentsProfileStatusRequestAndArrayAccessAsModel()
+    {
+        $action = new CancelRecurringPaymentsProfileAction();
+
+        $this->assertTrue(
+            $action->supports(new Cancel($this->getMock('ArrayAccess')))
```
makasim: There is no need to use mock here

Test two cases:

```
new Cancel(array('BILLINGPERIOD' => 'foo'));

//and

new Cancel(new \ArrayObject(array('BILLINGPERIOD' => 'foo')));
```

alexandernst:
@makasim \o/ Passing all the tests!

alexandernst:
@makasim Btw, do you want me to squash all those commits in one?

makasim:
>  Btw, do you want me to squash all those commits in one?

@alexandernst it would be good (:

makasim:
@alexandernst it would also good to merge it to 0.14 branch (there is not BC breaks), Could you do it?

alexandernst:
@makasim Ok, let me see how can I rebase that :)

alexandernst:
@makasim Done rebasing on top of 0.14 here https://github.com/Payum/Payum/pull/323

PR #324, opened 2015-03-12
Update obtain_checkout_token.html.twig

Perfect-Web:
should take into account the cents

### DIFF
--- a/src/Payum/Stripe/Resources/views/Action/obtain_checkout_token.html.twig
+++ b/src/Payum/Stripe/Resources/views/Action/obtain_checkout_token.html.twig
@@ -10,7 +10,7 @@
             data-image="{{ model.image|default("") }}"
             data-name="{{ model.name|default("") }}"
             data-description="{{ model.description|default("") }}"
-            data-amount="{{ model.amount }}">
+            data-amount="{{ model.amount*100 }}">
         </script>
     </form>
 {% endblock %}


makasim:
You have to set correct price

Perfect-Web:
i have set the correct price, but if i set the amount = $1000 on stripe is registered as $100.00
See http://stackoverflow.com/questions/25546372/stripe-payment-gateway-i-wants-to-use-amount-in-decimal-format

makasim:
I mean when you create payment details you have to pass it in cents (actually in format specific to a gateway you selected).

makasim:
Payum does not normalize data in any way (in case you dont use Order and FillOrderDetailsAction).

makasim:
see https://github.com/Payum/PayumBundle/blob/master/Resources/doc/custom_purchase_examples/stripe_js.md

there is an example of 1 USD purchase

Perfect-Web:
i figured that out, but when it is saved thru doctrine in db it will be saved as 100 not 1

makasim:
> i figured that out, but when it is saved thru doctrine in db it will be saved as 100 not 1

yes you are right

Perfect-Web:
so what do you suggest, cause as it is right now is very confusing ...

makasim:
what is confusing?

Perfect-Web:
that you set the amount you wanna charge to 100 times more than you wanna charge, and if you dont make any transformations when the process is done to the final amount, you will end up giving your clients 100 times more credit than you suppose to

makasim:
you have to store your price somewhere else in your own format, payment details is a payment gateway specific format and vary a lot, so do not use them directly,

PR #331, opened 2015-03-19
Update StatusAction.php

adelh45:
bundle was never seting the status on authorized


adelh45:
is it better here ?

makasim:
Thanks, I improved a bit the solution, please try the latest master branch and let me know weather it works for you or not.

PR #339, opened 2015-03-28
Add PDOStorage class

### LINKED ISSUE #205: [storage] a pdo storage
The storage takes a PDO connection, columns and can save\fetch a model to sql databases.

alexandernst:
Fixes #205


alexandernst:
@makasim What do you think? :)

alexandernst:
Ah, I forgot to mention it... for this to work the developer would need to create a table with 2 fields, 1 for the ID and 1 for the data itself. The data field could be any type of field that can handle a  serialized object (`longtext` would work just fine). As for the id field, a  `CHAR(13)` would do it.

makasim:
This is not generic enough to be accepted as it is now. For example you cannot use this storage with an Order model or a Token but it is a must. As I told the trickiest part is a schema which provides info on how to convert php object properties to database columns. And here we have to keep a balance. We should not reinvent one more complete ORM abstractions, it should be simple but provide basic features.

src/Payum/Core/Storage/PDOStorage.php
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * @param string $dsn
+     * @param string $user
+     * @param string $pass
+     * @param string $table
+     * @param string $idkey
+     * @param array $options
+     */
+    public function __construct($dsn, $user, $pass, $table, $idkey, $options = array())
```
makasim: The argument has to be an instance of already created PDO.

src/Payum/Core/Storage/PDOStorage.php
```diff
@@ -0,0 +1,168 @@
...
+     */
+    private function _insertOrUpdateModel($id, $model)
+    {
+        $data = $this->_queryModel($id);
+        $model = serialize($model);
+
+        if($data !== false){
+            $this->dbh->exec("UPDATE `payum-storage` SET `payum-data-model` = '$model' WHERE `$this->idkey` = '$id'");
+        }else{
+            $this->dbh->exec("INSERT INTO `payum-storage` (`$this->idkey`, `payum-data-model`) VALUES ('$id', '$model')");
```
makasim: table name has to be configurable.

makasim:
We can have a schema like 

```php
<?php

$schema = array(
  array('column' => 'aDbColumn', 'property' => 'aObjProperty', 'toDb' => 'callable', 'toPhp' => 'callable', 'primary' => true);
);
```

And for example for a model with two properties id and serialized the schema will look like:

```php
<?php

$schema = array(
  array('column' => 'id', 'property' => 'id', 'primary' => true),
  array('column' => 'details', 'property' => 'details', 'toDb' => 'json_encode', 'toPhp' => 'json_decode'),
);

$storage = new PDOStorage($pdo, $schema, $class);
```

src/Payum/Core/Storage/PDOStorage.php
```diff
@@ -0,0 +1,168 @@
+<?php
+namespace Payum\Core\Storage;
```
makasim: Could you please move it to Payum\Bridge\PlainPhp\Storage namespace?

alexandernst:
@makasim Patched the code with your suggestions, except the schema one. I'm not really sure I quite understand it. How many different object is the storage class supposed to store? And also, how does the FileStorage class is able to handle them without all that fuzz?

src/Payum/Core/Bridge/PlainPhp/Storage/PDOStorage.php
```diff
@@ -0,0 +1,161 @@
...
+     * @var array
+     */
+    protected $identityMap;
+
+    /**
+     * @param object $pdo
+     * @param string $table
+     * @param string $idkey
+     */
+    public function __construct($pdo, $table, $idkey)
```
makasim: typehint for $pdo argument

makasim:
> How many different object is the storage class supposed to store?

ideally anything a developer want to. but at least:

* https://github.com/Payum/Payum/blob/master/src/Payum/Core/Model/Token.php
* https://github.com/Payum/Payum/blob/master/src/Payum/Core/Model/ArrayObject.php
* https://github.com/Payum/Payum/blob/master/src/Payum/Core/Model/Order.php

But we do a schema approach all of them will be converted.

makasim:
> And also, how does the FileStorage class is able to handle them without all that fuzz?

it just [serialize](https://github.com/Payum/Payum/blob/master/src/Payum/Core/Storage/FilesystemStorage.php#L81) the object to a file, which is not the case for database.

alexandernst:
@makasim I haven't given up on this one, just no free time :(

Anyways, just to keep the conversation up, why the PDO storage class can't handle the data the same way the File storage class does? I mean, in the File storage class the file name is used as an ID, right? And the content of the file is the serialized data.

In a SQL storage, a column `_id` could be used as an ID, and another column of the type of `VARCHAR` of  `TEXT` can be used to store the serialized data. Can't it be done that way?

makasim:
It could be done indeed that way. But it is not useful, since we are not
using advantages of databases. Filesystem storage is done that way to keep
things simple and because of the limitations of filtered. We don't have
such in case of database

Пн, 6 апр. 2015 г. в 18:02, Alexander Nestorov <notifications@github.com>:

harikt:
I am interested to give this a try. I will see if I can make some time for this.

PR #366, opened 2015-04-29
RenderTemplate request additional parameters

serdyuka:


serdyuka:
@makasim review please

src/Payum/Core/Request/RenderTemplate.php
```diff
@@ -60,4 +60,13 @@ public function setResult($result)
     {
         $this->result = $result;
     }
+
+    /**
+     * @param string $name
+     * @param mixed  $value
+     */
+    public function addParameter($name, $value)
```
makasim: let's add one more merhod `setParamter`. and this one must throw exception if there is already such parameter set where `setParameter` should always set\overwrite it. 

Could you rename context to `parameters` and gettter too.

src/Payum/Stripe/Resources/views/Action/obtain_js_token.html.twig
```diff
@@ -5,7 +5,7 @@
 {% block payum_body %}
     {{ parent() }}
 
-    <form action="" method="POST" id="payment-form">
+    <form action="{{ captureActionUrl|default('') }}" method="POST" id="payment-form">
```
makasim: name it `actionUrl`

serdyuka:
@makasim done

makasim:
Thanks

PR #367, opened 2015-04-30
Klarna Invoice Update

ASKozienko:


src/Payum/Klarna/Invoice/Action/Api/UpdateAction.php
```diff
@@ -0,0 +1,59 @@
...
+        RequestNotSupportedException::assertSupports($this, $request);
+
+        $details = ArrayObject::ensureArrayObject($request->getModel());
+
+        $klarna = $this->getKlarna();
+
+        try {
+            $this->gateway->execute(new PopulateKlarnaFromDetails($details, $klarna));
+
+            $details['updated'] = $klarna->update($details['rno']);
```
makasim: @ASKozienko what if order is updated several times? shouldn't it be an array like date -> status 

just thoughts

makasim:
lgtm

PR #369, opened 2015-05-01
use Stripe js object after script load

serdyuka:


serdyuka:
@makasim ready for review

src/Payum/Stripe/Resources/views/Action/obtain_js_token.html.twig
```diff
@@ -45,26 +45,6 @@
     <script type="text/javascript" src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>
```
makasim: could you create a separate block payum_vendor_javascripts and move these to it.

src/Payum/Stripe/Resources/views/Action/obtain_js_token.html.twig
```diff
@@ -72,11 +52,31 @@
                 // Disable the submit button to prevent repeated clicks
                 $form.find('button').prop('disabled', true);
 
+                // This identifies your website in the createToken call below
+                Stripe.setPublishableKey({{ publishable_key|json_encode|raw }});
```
makasim: not sure it must be in submit.

makasim:
@serdyuka ping

serdyuka:
@makasim done

serdyuka:
@makasim done

makasim:
Thanks.

PR #377, opened 2015-05-21
Added currency to checkout token template

liquorvicar:
This is my first contribution to payum and I couldn't find any contribution guidelines so let me know if there's any styling issues or extra tests etc I need to add.

### DIFF
--- a/src/Payum/Stripe/Resources/views/Action/obtain_checkout_token.html.twig
+++ b/src/Payum/Stripe/Resources/views/Action/obtain_checkout_token.html.twig
@@ -10,7 +10,8 @@
             data-image="{{ model.image|default("") }}"
             data-name="{{ model.name|default("") }}"
             data-description="{{ model.description|default("") }}"
-            data-amount="{{ model.amount }}">
+            data-amount="{{ model.amount }}"
+            data-currency="{{ model.currency }}">
         </script>
     </form>
 {% endblock %}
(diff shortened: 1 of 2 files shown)


src/Payum/Stripe/Resources/views/Action/obtain_checkout_token.html.twig
```diff
@@ -10,7 +10,8 @@
             data-image="{{ model.image|default("") }}"
             data-name="{{ model.name|default("") }}"
             data-description="{{ model.description|default("") }}"
-            data-amount="{{ model.amount }}">
+            data-amount="{{ model.amount }}"
+            data-currency="{{ model.currency }}">
```
makasim: looks good, shouldn't we set usd as default here?

liquorvicar: Good call. I wasn't sure whether the currency would always be set on the model or not but adding the default gets round that (and doesn't break existing behaviour). Have amended pull request.

makasim:
Thanks.

liquorvicar:
@makasim Thanks, any chance you can tag this as a release?

makasim:
@liquorvicar I do not plan to release 1.0 in near future cuz there are some  issues which must be included it. What we can do is to backport the fix to 0.15 branch. It is stable one and I can tag at any moment

liquorvicar:
Adding it to the 0.x version would be absolutely fine. Thanks!

Issue #250, opened 2014-11-20
Paypal and TLSv1


smarques:
Hi  one question... I have seen the changes on https://github.com/formapro-forks/Payum/commit/08a9c986afd1a41cc63f339810f8a4c4353a312b have been incorporated into the main repo.
But as far I can see it has not been integrated payum/paypal-express-checkout-nvp
So how do I go about making paypal work again?

Thank you for any help.
sergio

naabys:
Updating the payum libs didn't did the trick for me. I checked the latest version of paypal php sdk and found that they remove `CURLOPT_SSL_CIPHER_LIST` from the client options if they find NSS instead of OpenSSL [see here](https://github.com/paypal/sdk-core-php/blob/namespace-5.3/lib/PayPal/Core/PPHttpConfig.php#L51). 
As far as I'm concerned, I overrode the `payum.buzz.client` service to set the option to null in BuzzClient:
```
  payum.buzz.client:
    class: Buzz\Client\ClientInterface
    factory_class: Payum\Core\Bridge\Buzz\ClientFactory
    factory_method: createCurl
    calls:
      - [setTimeout,[200]]
      # remove CURLOPT_SSL_CIPHER_LIST from options list
      - [setOption, [10083, null]]
```

hope this helps

makasim:
@naabys thanks for the info, I'll try to learn more about it.

Don87:
@naabys @smarques I'm a little bit of a noob here, which file exactly do I edit/create and paste that information in? I am using latest Symfony and Payum update vendors... three times to be sure, cleared cache but still get this error on Paypal Express Checkout 

Unknown cipher in list: TLSv1

Anyone, please help coz I am well and truly stuck.

smarques:
@Don87 just copy naabys solution somewhere in your configuration files, under the configuration block
services,
for example in app/config/config.yml


    services:
       payum.buzz.client:
          class: Buzz\Client\ClientInterface
          factory_class: Payum\Core\Bridge\Buzz\ClientFactory
          factory_method: createCurl
          calls:
            - [setTimeout,[200]]
            # remove CURLOPT_SSL_CIPHER_LIST from options list
            - [setOption, [10083, null]]

Don87:
[...]

PR #399, opened 2015-08-02
Do not use the Cipher List for NSS

StudioMaX:
According to this issues:
https://github.com/Payum/Payum/issues/250
https://github.com/Payum/PayumBundle/issues/294

I know that in the current master branch Buzz replaced by the Guzzle PSR-7, but a lot of projects uses stable branch. Could you fix this issue in the current stable version?

### DIFF
--- a/src/Payum/Core/Bridge/Buzz/ClientFactory.php
+++ b/src/Payum/Core/Bridge/Buzz/ClientFactory.php
@@ -15,7 +15,12 @@ class ClientFactory
         //reaction to the ssl3.0 shutdown from paypal
         //https://www.paypal-community.com/t5/PayPal-Forward/PayPal-Response-to-SSL-3-0-Vulnerability-aka-POODLE/ba-p/891829
         //http://googleonlinesecurity.blogspot.com/2014/10/this-poodle-bites-exploiting-ssl-30.html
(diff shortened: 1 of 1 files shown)


makasim:
Could you provide more details on issue that you have? Why is this change required?

StudioMaX:
TLSv1 does not exists in the old versions of NSS (i have 7.19.7, which is used on CentOS 6.6).
When we call curl with the CURLOPT_SSL_CIPHER_LIST, TLSv1 it raises an error "Unknown cipher in list: TLSv1". In this regard, it is impossible to use PayPal with Payum.

makasim:
Does paypal works with your fix?

StudioMaX:
Yes, now it works. This fix also used in the paypal sdk
https://github.com/paypal/sdk-core-php/blob/namespace-5.3/lib/PayPal/Core/PPHttpConfig.php#L51

makasim:
merged as part of https://github.com/Payum/Payum/pull/400

makasim:
0.15.3 contains the fix.

makasim:
@StudioMaX could you please test it?

StudioMaX:
Thanks! It works now on 0.15.3.

PR #401, opened 2015-08-03
Pass model with obtain credit card request

makasim:
Fixes https://github.com/Payum/Payum/issues/397


makasim:
ping @antonioperic

antonioperic:
This is nice :+1:

PR #405, opened 2015-09-10
Update supported-gateways.md

fullpipe:
Add Uniteller and Flexidengi payment gateways

### DIFF
--- a/src/Payum/Core/Resources/docs/supported-gateways.md
+++ b/src/Payum/Core/Resources/docs/supported-gateways.md
@@ -30,6 +30,8 @@
 * [peterfox/Payum-Bitpay](https://github.com/peterfox/Payum-Bitpay)
 * [winzou/payum-limonetik](https://github.com/winzou/payum-limonetik)
 * [ekipower/payum-nganluong](https://github.com/ekipower/payum-nganluong)
+* [fullpipe/payum-uniteller](https://github.com/fullpipe/payum-uniteller)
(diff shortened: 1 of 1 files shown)


makasim:
Great! Thanks!

One suggestion so far. You can reuse [unsafe notify action](https://github.com/Payum/PayumBundle/blob/master/Controller/NotifyController.php#L10). And implement the logic like this  https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/PayumBundle/Payum/Be2bill/Action/NotifyAction.php

makasim:
Also you can add an extension which looks for notify requests with expected model and checks its status. if status is success you dispatch an event. If you dont want bother with it you can use build it extension: https://github.com/Payum/Payum/blob/master/src/Payum/Core/Bridge/Symfony/Extension/EventDispatcherExtension.php

makasim:
Or like this https://github.com/Payum/Payum/pull/408

fullpipe:
@makasim what do you think about adding `StorageAwareInterface` for actions just like `ApiAwareInterface`. For example for `GetPaymentFromHttpRequestRequest($_REQUEST)` 

And add something like `GatewayRequestGueser::getGateway($_REQUEST)` for using with unsafe notifications to identify gateway from request params.

makasim:
@fullpipe  Not sure it has to be in the lib itself, Rare case, only few gateways does not support notify url per payment and also it is easy to implement.

PR #417, opened 2015-10-18
Fixed ajbdev/authorizenet-php-api deprecation

herodrigues:
Fixed the authorizenet deprecation and update to its latest version.


composer.json
```diff
@@ -58,7 +58,7 @@
         "symfony/http-foundation": "~2.3",
         "symfony/form": "~2.3",
         "symfony/validator": "~2.6",
-        "ajbdev/authorizenet-php-api": "1.1.*",
+        "authorizenet/authorizenet": "1.8.*",
```
makasim: It would be good to set it like `~1.8`

makasim: You have to update this line https://github.com/Payum/Payum/pull/417/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780L93 too, and the composer.json of Payum/AuthorizeNet/Aim component

makasim:
@herodrigues Thanks for the PR, does it work for  you?

herodrigues:
Ok, I'm gonna do that

herodrigues:
It's done :+1:

makasim:
@herodrigues merged, thanks for PR

PR #443, opened 2015-12-10
Improve serialize / unserialize when using FilesystemStrorage

daveli:
I encountered an issue when dealing with serialize and unserialize some "unsafe" (", ', :, or ; in the) data. This patch will fix that.

### DIFF
--- a/src/Payum/Core/Storage/FilesystemStorage.php
+++ b/src/Payum/Core/Storage/FilesystemStorage.php
@@ -52,7 +52,13 @@ class FilesystemStorage extends AbstractStorage
         }
 
         if (file_exists($this->storageDir.'/payum-model-'.$id)) {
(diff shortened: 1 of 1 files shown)


makasim:
What was the problem?

makasim:
FYI: you should not use this store in prod.

makasim:
Its goal to allow you easily and quickly start with Payum ,you  have to replace it with something else ASAP.

daveli:
This is pure a patch for testing purposes. I just happen to ran into it and it was a quick fix. I'm using a model at this very moment.

makasim:
What did you get, an exception or error, corrupted data?

makasim:
The fix should be done in no BC break way. which is not how it is right now.

daveli:
Typical corrupted data with serialize: Error at offset x of x byte in file x.php

PR #447, opened 2015-12-16
Add Sofort gateway

makasim:
The feature is blocked by Sofort. They are not able to provide me with access to sandbox.


makasim:
close in favor of https://github.com/Payum/Payum/pull/477. same code but rebased.

PR #458, opened 2016-01-14
Add symfony 3.0 support

makasim:


makasim:
It fails because of omnipay-common has hard dependency on symfony 2.x package. It has to be fixed before we can processed here.


https://github.com/thephpleague/omnipay-common/pull/73

PR #459, opened 2016-01-19
Bump Stripe API to version 2.1.3 [WIP]

maddo:
Investigating updating the Stripe API version in Payum

## End Objective
Latest Stripe API in Payum 1.x

## First step
Testing 2.1.3 in .14

## Next steps
Either 2.1.3 in Payum 1.x or directly latest Stripe in Payum 1.x

## Why 2.1.3?
`Network requests are now done through a swappable class for easier mocking` https://github.com/stripe/stripe-php/blob/v2.1.3/CHANGELOG.md

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
         "paypal/rest-api-sdk-php" : "0.5.*",
         "klarna/checkout": "~1",
         "fp/klarna-invoice": "0.1.*",
-        "stripe/stripe-php": "~1.0",
+        "stripe/stripe-php": "2.1.3",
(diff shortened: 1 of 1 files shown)


maddo:
Changed branch name

makasim:
I'd really like to migrate to stripe 2.x version but it is  BC break since version one and version two are not compatible. At least class names are different, there are namespaces in version two.

I have to think of an impact this change may have.

maddo:
Thanks! This was a silly PR. I have another one incoming that addresses the namespace changes.

PR #461, opened 2016-01-20
Updates stripe api to 3.6

maddo:
Includes namespace changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,7 +63,7 @@
         "paypal/rest-api-sdk-php" : "0.5.*",
         "klarna/checkout": "~1|~2.0",
         "fp/klarna-invoice": "0.1.*",
-        "stripe/stripe-php": "~1.0",
+        "stripe/stripe-php": "~2.0|~3.0",
         "ext-soap": "*"
     },
     "replace": {
@@ -95,7 +95,7 @@
         "paypal/rest-api-sdk-php" : "0.5.* If you want to use PayPal REST API install paypal/rest-api-sdk-php:0.5.* library",
         "klarna/checkout": "~1.0|~2.0 If you want to use Klarna Checkout install klarna/checkout:~1|~2.0 library",
         "fp/klarna-invoice": "0.1.* If you want to use Klarna Invoice install fp/klarna-invoice:0.1.* library",
-        "stripe/stripe-php": "~1.0 If you want to use Stripe install stripe/stripe-php:~1.0 library",
+        "stripe/stripe-php": "~2.0|~3.0 If you want to use Stripe install stripe/stripe-php:~2.0|~3.0 library",
         "ext-soap": "* If you want to use Payex extension you have to install soap extension"
     },
     "config": {
--- a/src/Payum/Stripe/StripeCheckoutGatewayFactory.php
+++ b/src/Payum/Stripe/StripeCheckoutGatewayFactory.php
@@ -16,8 +16,8 @@ class StripeCheckoutGatewayFactory extends GatewayFactory
      */
     protected function populateConfig(ArrayObject $config)
     {
-        if (!class_exists('Stripe')) {
-            throw new \LogicException('You must install "stripe/stripe-php" library.');
+        if (!class_exists('\Stripe\Stripe')) {
+            throw new \LogicException('You must install "stripe/stripe-php:~2.0|~3.0" library.');
         }
 
         $config->defaults(array(
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrades
 
+## 1.2.0
+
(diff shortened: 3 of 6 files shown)


src/Payum/Stripe/StripeCheckoutGatewayFactory.php
```diff
@@ -16,7 +16,7 @@ class StripeCheckoutGatewayFactory extends GatewayFactory
      */
     protected function populateConfig(ArrayObject $config)
     {
-        if (!class_exists('Stripe')) {
```
makasim: if user installed the old version, this where this class Stripe exists we have to throw an exception. We have to told a user that the minimum required version of stripe is 3.6 and he has to upgrade it too.

src/Payum/Stripe/StripeCheckoutGatewayFactory.php
```diff
@@ -16,7 +16,7 @@ class StripeCheckoutGatewayFactory extends GatewayFactory
...
+        if (!class_exists('\Stripe\Stripe')) {
             throw new \LogicException('You must install "stripe/stripe-php" library.');
```
makasim: I am thinking of adding a version here. So it is clear what version of the library user has to install. wdyt?

maddo: Don't see why not :)
Also, I'm pretty sure that there are not breaking changes from 2.0 - 3.6, for this limited usage. So there is no need to make this so tightly coupled to 3.6. Updating to support that range. (Namespaces were introduced in 2.0)

makasim: So maybe we can set `~2.0|~3.0` as supported versions?

maddo: I suppose so, they seem to respect semver. Not sure if it's better to be cautious and only support up to the current version.

src/Payum/Stripe/StripeCheckoutGatewayFactory.php
maddo: Installed dependencies on master, then switched to my branch and ran phpunit. This seems about right: 
```
/code/Payum/Payum/src/Payum/Stripe on bump-stripe λ phpunit
PHPUnit 4.8.21 by Sebastian Bergmann and contributors.

...............................................................  63 / 128 ( 49%)
.............................................EEEEEEFE....EEEEEE 126 / 128 ( 98%)
FE

Time: 163 ms, Memory: 11.00Mb

There were 14 errors:

1) Payum\Stripe\Tests\StripeCheckoutGatewayFactoryTest::shouldAllowCreateGateway
LogicException: You must install "stripe/stripe-php":"2.0 - 3.6" library.

/code/Payum/Payum/src/Payum/Stripe/StripeCheckoutGatewayFactory.php:20
/code/Payum/Payum/src/Payum/Stripe/vendor/payum/core/Payum/Core/GatewayFactory.php:45
/code/Payum/Payum/src/Payum/Stripe/vendor/payum/core/Payum/Core/GatewayFactory.php:33
/code/Payum/Payum/src/Payum/Stripe/Tests/StripeCheckoutGatewayFactoryTest.php:55
/usr/share/php/PHPUnit/TextUI/Command.php:147
/usr/share/php/PHPUnit/TextUI/Command.php:99
...
```

composer.json
```diff
@@ -95,7 +95,7 @@
         "paypal/rest-api-sdk-php" : "0.5.* If you want to use PayPal REST API install paypal/rest-api-sdk-php:0.5.* library",
         "klarna/checkout": "~1.0|~2.0 If you want to use Klarna Checkout install klarna/checkout:~1|~2.0 library",
         "fp/klarna-invoice": "0.1.* If you want to use Klarna Invoice install fp/klarna-invoice:0.1.* library",
-        "stripe/stripe-php": "~1.0 If you want to use Stripe install stripe/stripe-php:~1.0 library",
+        "stripe/stripe-php": "~3.6 If you want to use Stripe install stripe/stripe-php:~3.6 library",
```
maddo: Left the hint at the latest version in the range.

src/Payum/Stripe/StripeCheckoutGatewayFactory.php
makasim: replace "stripe/stripe-php":"2.0 - 3.6" with "stripe/stripe-php:~2.0|~3.0" so a user can copy and past it to composer console  command.

composer.json
makasim: ok

makasim:
Last thing. Could you please add a note to UPGRADE.md. This change will be in 1.2.0 release.

makasim:
Thanks for contributing it. It will be released in 1.2.0 version. Though I have to test in the sandbox first.

maddo:
:+1: thanks for the quick feedback!

PR #489, opened 2016-02-29
Fixes incomplete annotations of magic methods

ZloeSabo:
According the [docs](https://phpdoc.org/docs/latest/references/phpdoc/tags/method.html) method definition should contain brackets. Omitting those breaks any test suite which depends on Prophecy, when trying to double class, inherited from interface with magic method.

Example can be found [here](https://gist.github.com/ZloeSabo/5c9233753d0c8c493c90)


makasim:
Thanks.

PR #497, opened 2016-03-16
[WIP] Add BankAccount

davidfuhr:
To support Direct debit payment (https://en.wikipedia.org/wiki/Direct_debit) or Giropay (https://en.wikipedia.org/wiki/Giropay) with different payment providers it is very useful to have a common interface for the bank account.

The Paypal API has way more fields for the bank account (https://developer.paypal.com/docs/classic/api/adaptive-accounts/AddBankAccount_API_Operation/). But I think it makes no sense to add all those fields.

@makasim what are your thoughts about that?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.2.4 (2016-03-07)
+
+* [credit-card] Fixes **regression** in the bundle. Symfony's validator marks all fields as invalid on credit card form submission.
+* [skeleton] fix some typos.
+
 ## 1.2.3 (2016-03-04)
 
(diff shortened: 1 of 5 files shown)


makasim:
Are you sure these properties (holder, number etc) cover all cases?

makasim:
> But I think it makes no sense to add all those fields.

Yes sure, they could be set via payment->setDetails method (gateway specific format). I am thinking whether we need such a method on bank account obj.

makasim:
@davidfuhr do you have something implemented on top of this?

makasim:
Merged but some logic marked as experimental. It means that those classes could be changed at any moment.

makasim:
@davidfuhr thanks.

PR #505, opened 2016-03-21
Using Httplug instead of Guzzle6

### LINKED ISSUE #491: Replace Guzzle6 with HTTPlug
I order to remove the dependency on Guzzle I suggest we use [Httplug](http://docs.php-http.org/en/latest/) instead. This way we can let our users to still use Buzz, Guzzle5, Curl or whatever they want. 

Im happy to send a PR if you think this idea is worth considering.

Nyholm:
This is PR will remove our hard coupled dependency on Guzzle6 and instead rely on an interface. This will make the Payum package more SOLID. With this PR we allow users that are stuck with using Guzzle5 to use Payum. 

~~This PR introduce some small BC breaks. I've removed the function `HttpClientFactory::createGuzzle` and also the config parameter named `guzzle.client`.~~

This will fix #491

### Usage

```php
/** @var Payum $payum */
$payum = (new PayumBuilder())
    ->addDefaultStorages()
    ->addGateway('gatewayName', [
        'factory' => 'aGateway'
        'httplug.client' => new \Http\Adapter\Guzzle6\Client(),
    ])
    ->getPayum()
;
```

### TODO
* [x] Make code changes
* [x] Deprecate existing classes tightly coupled to Guzzle
* [x] Add documentation

### Implications

When you update to Payum 1.3.0 the installation will fail because you need to install a [client implementation](https://packagist.org/providers/php-http/client-implementation). If you choose `php-http/guzzle6-adapter` everything will work **exactly** as before. No BC breaks.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,7 +2,7 @@
 
 ## 1.3.0
 
-* [http-client] guzzlehttp/guzzle was decoupled and replaced by [Httplug](http://docs.php-http.org/en/latest/httplug/introduction.html). This gives you option to use Guzzle6, Guzzle5, Buzz or any other library to send HTTP messages. Make sure to specify `httplug.client` and `httplug.message_factory` or use [Httplug discovery](http://docs.php-http.org/en/latest/discovery.html). 
+* [http-client] When you update to Payum 1.3.0 the installation will fail because you need to install a client implementation. If you choose php-http/guzzle6-adapter everything will work exactly as before.
 
 ## 1.2.0
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrades
 
+## 1.3.0
+
+* [http-client] guzzlehttp/guzzle was decoupled and replaced by [Httplug](http://docs.php-http.org/en/latest/httplug/introduction.html). This gives you option to use Guzzle6, Guzzle5, Buzz or any other library to send HTTP messages. Make sure to specify `httplug.client` and `httplug.message_factory` or use [Httplug discovery](http://docs.php-http.org/en/latest/discovery.html). 
+
 ## 1.2.0
 
 * [stripe] Stripe api version updated to 2.0 - 3.x. You'll have to update any custom actions that use the Stripe api directly.
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,10 @@
     "require": {
         "php": "^5.5.0|^7.0",
         "payum/iso4217": "~1.0",
-        "guzzlehttp/guzzle": "~6.0",
+        "php-http/httplug": "^1.0",
+        "php-http/message-factory": "^1.0",
+        "php-http/client-implementation": "^1.0",
+        "php-http/discovery": "^0.8",
         "league/url": "~3.0",
         "twig/twig": "~1.0"
     },
@@ -66,6 +69,10 @@
         "klarna/checkout": "~1|~2.0",
         "fp/klarna-invoice": "0.1.*",
         "stripe/stripe-php": "~2.0|~3.0",
+        "php-http/guzzle6-adapter": "^1.0",
+        "php-http/message": "^1.0",
+        "puli/composer-plugin": "^1.0",
+        "guzzlehttp/psr7": "^1.1",
         "ext-soap": "*"
     },
     "replace": {
@@ -102,6 +109,7 @@
         "stripe/stripe-php": "~2.0|~3.0 If you want to use Stripe install stripe/stripe-php:~2.0|~3.0 library",
         "ext-soap": "* If you want to use Payex extension you have to install soap extension"
     },
+    "minimum-stability": "beta",
     "config": {
         "bin-dir": "bin"
     },
--- a/src/Payum/Be2Bill/Api.php
+++ b/src/Payum/Be2Bill/Api.php
@@ -1,9 +1,7 @@
 <?php
 namespace Payum\Be2Bill;
 
-use Http\Discovery\MessageFactoryDiscovery;
 use Http\Message\MessageFactory;
-use Payum\Core\Bridge\Guzzle\HttpClientFactory;
 use Payum\Core\Bridge\Spl\ArrayObject;
 use Payum\Core\Exception\InvalidArgumentException;
 use Payum\Core\Exception\Http\HttpException;
@@ -130,7 +128,7 @@ class Api
      *
      * @throws \Payum\Core\Exception\InvalidArgumentException if an option is invalid
      */
-    public function __construct(array $options, HttpClientInterface $client = null, MessageFactory $messageFactory = null)
+    public function __construct(array $options, HttpClientInterface $client, MessageFactory $messageFactory )
     {
         $options = ArrayObject::ensureArrayObject($options);
         $options->defaults($this->options);
@@ -144,8 +142,8 @@ class Api
         }
 
         $this->options = $options;
-        $this->client = $client ?: HttpClientFactory::create();
-        $this->messageFactory = $messageFactory ?: MessageFactoryDiscovery::find();
+        $this->client = $client;
+        $this->messageFactory = $messageFactory;
     }
 
     /**
--- a/src/Payum/Be2Bill/Api.php
+++ b/src/Payum/Be2Bill/Api.php
@@ -1,7 +1,8 @@
 <?php
 namespace Payum\Be2Bill;
 
-use GuzzleHttp\Psr7\Request;
+use Http\Discovery\MessageFactoryDiscovery;
+use Http\Message\MessageFactory;
 use Payum\Core\Bridge\Guzzle\HttpClientFactory;
 use Payum\Core\Bridge\Spl\ArrayObject;
 use Payum\Core\Exception\InvalidArgumentException;
@@ -109,6 +110,11 @@ class Api
     protected $client;
 
     /**
+     * @var MessageFactory
+     */
+    protected $messageFactory;
+
+    /**
      * @var array
      */
     protected $options = array(
@@ -120,10 +126,11 @@ class Api
     /**
      * @param array               $options
      * @param HttpClientInterface $client
+     * @param MessageFactory      $messageFactory
      *
      * @throws \Payum\Core\Exception\InvalidArgumentException if an option is invalid
      */
-    public function __construct(array $options, HttpClientInterface $client = null)
+    public function __construct(array $options, HttpClientInterface $client = null, MessageFactory $messageFactory = null)
     {
         $options = ArrayObject::ensureArrayObject($options);
         $options->defaults($this->options);
@@ -138,6 +145,7 @@ class Api
 
         $this->options = $options;
         $this->client = $client ?: HttpClientFactory::create();
+        $this->messageFactory = $messageFactory ?: MessageFactoryDiscovery::find();
     }
 
     /**
@@ -187,7 +195,7 @@ class Api
             'Content-Type' => 'application/x-www-form-urlencoded',
         );
 
-        $request = new Request('POST', $this->getApiEndpoint(), $headers, http_build_query($fields));
+        $request = $this->messageFactory->createRequest('POST', $this->getApiEndpoint(), $headers, http_build_query($fields));
 
         $response = $this->client->send($request);
 
--- a/src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
+++ b/src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
@@ -14,7 +14,7 @@ class HttpClientFactory
 {
     /**
      * Create a Guzzle client.
-     *
+     * 
      * @return \GuzzleHttp\Client
      */
     public static function createGuzzle()
--- a/src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
+++ b/src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
@@ -1,26 +1,43 @@
 <?php
+
 namespace Payum\Core\Bridge\Guzzle;
 
-use GuzzleHttp\Client as GuzzleClient;
-use GuzzleHttp\ClientInterface as GuzzleClientInterface;
+use GuzzleHttp\Client;
+use Http\Discovery\HttpClientDiscovery;
 use Payum\Core\HttpClientInterface;
 
+/**
+ * @deprecated This will be removed in 2.0. Consider using Http\Discovery\HttpClientDiscovery.
+ */
 class HttpClientFactory
 {
     /**
-     * @return GuzzleClientInterface
+     * Create a Guzzle client.
+     *
+     * @return \GuzzleHttp\Client
      */
     public static function createGuzzle()
     {
+        $client = null;
+        if (!class_exists(Client::class)) {
+            @trigger_error('The function "HttpClientFactory::createGuzzle" is depcrecated and will be removed in 2.0.', E_USER_DEPRECATED);
+            throw new \LogicException('Can not use "HttpClientFactory::createGuzzle" since Guzzle is not installed. This function is deprecated and will be removed in 2.0.');
+        }
+
+        $version = \GuzzleHttp\ClientInterface::VERSION;
+        if (substr($version, 0, 1) !== '6') {
+            throw new \LogicException('This version of Guzzle is not supported.');
+        }
+
         $curl = curl_version();
 
         $curlOptions = [
-            CURLOPT_USERAGENT => sprintf('Payum/1.x curl/%s PHP/%s', $curl['version'], phpversion()),
-        ];
+             CURLOPT_USERAGENT => sprintf('Payum/1.x curl/%s PHP/%s', $curl['version'], phpversion()),
+         ];
 
-        return new GuzzleClient([
-            'curl' => $curlOptions
-        ]);
+        return new \GuzzleHttp\Client([
+             'curl' => $curlOptions,
+         ]);
     }
 
     /**
@@ -28,6 +45,6 @@ class HttpClientFactory
      */
     public static function create()
     {
-        return new HttpClient(static::createGuzzle());
+        return new HttplugClient(HttpClientDiscovery::find());
     }
 }
--- a/src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
+++ b/src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
@@ -4,7 +4,6 @@ namespace Payum\Core\Bridge\Guzzle;
 
 use GuzzleHttp\Client;
 use Http\Discovery\HttpClientDiscovery;
-use Payum\Core\Bridge\Httplug\HttplugClient;
 use Payum\Core\HttpClientInterface;
 
 /**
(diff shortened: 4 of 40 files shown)


src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
```diff
@@ -3,31 +3,20 @@
...
-        ]);
-    }
-
-    /**
      * @return HttpClientInterface
      */
     public static function create()
     {
-        return new HttpClient(static::createGuzzle());
+        return new HttpClient(HttpClientDiscovery::find());
```
makasim: it throws an exception if there is not any implementations (packages), am I right?

composer.json
```diff
@@ -39,7 +39,9 @@
     "require": {
         "php": "^5.5.0|^7.0",
         "payum/iso4217": "~1.0",
-        "guzzlehttp/guzzle": "~6.0",
+        "php-http/client-implementation": "^1.0",
+        "php-http/discovery": "^0.8",
+        "guzzlehttp/psr7": "^1.1",
```
makasim: this is only for BC? right?

src/Payum/Core/CoreGatewayFactory.php
```diff
@@ -98,6 +97,11 @@ public function createConfig(array $config = array())
             };
         }
 
+        if (empty($config['payum.http_client'])) {
+            $httplugClient = $config['httplug_client'] ?: HttpClientDiscovery::find();
+            $config['payum.http_client'] = new HttpClient($httplugClient);
+        }
+
```
makasim: Looks better to me:

```php
<?php
$config->default([
    'payum.http_client' => function(ArrayObject $config) {
        return new HttpClient($config['httplug_client']);
    },
    'httplug_client' => function(ArrayObject $config) {
        return HttpClientDiscovery::find();
    }
]);
```

composer.json
makasim: Why should it be required explicitly?

composer.json
```diff
@@ -102,6 +106,7 @@
         "stripe/stripe-php": "~2.0|~3.0 If you want to use Stripe install stripe/stripe-php:~2.0|~3.0 library",
         "ext-soap": "* If you want to use Payex extension you have to install soap extension"
     },
+    "minimum-stability": "beta",
```
makasim: why setting beta globally?

Nyholm: Puli is not stable yet. Either this or require 4 extra packages that are in beta.

composer.json
Nyholm: We are still using the psr7 implementation of guzzle. Not the http client. 
I could remove this dependency as well if you want.

src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
Nyholm: Yes. If you try to use discovery without installing puli or if you do not have a psr7 implementation, this will throw an exception.

src/Payum/Core/CoreGatewayFactory.php
Nyholm: Agreed. I'll change this.

Nyholm:
Thank you for reviewing this so quickly.

composer.json
makasim: ok

composer.json
makasim: @Nyholm As soon as composer fails on update when there is not an implementation for `php-http/client-implementation` we are safe and can get rid of guzzle as default package. Is this how it expected to work?

Nyholm: Since we have required a virtual package (`php-http/client-implementation`) composer will fail if we are missing a package that provides `php-http/client-implementation`. The virtual package guarantees **an** implementation of a php-http client but not which one. You can consider a virtual package like an interface. 

Guzzle has two packages. One PSR7 implementation and one HttpClient. I have removed the dependency of the HttpClient. We are still dependent on the PSR7 implementation. I could do some work and remove the dependency of the PSR7 implementation as well.

src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
makasim: So we can drop a guzzle as default client, since users are forced to take care of this on composer update.

composer.json
makasim: > We are still dependent on the PSR7 implementation. I c

Is it somewhere In the Payum code or in Httplug?

src/Payum/Core/CoreGatewayFactory.php
```diff
@@ -111,7 +116,7 @@ public function createConfig(array $config = array())
     protected function buildClosures(ArrayObject $config)
     {
         // with higher priority
-        foreach (['guzzle.client', 'payum.http_client', 'payum.paths', 'twig.env'] as $name) {
+        foreach (['payum.http_client', 'payum.paths', 'twig.env'] as $name) {
```
makasim: @Nyholm I think you have to add a `httplug_client` options here too.

composer.json
Nyholm: Payum. Every time you [create a Request](https://github.com/Payum/Payum/blob/master/src/Payum/Skeleton/Api.php#L42) you use the Guzzle implementation. 
I could use Httplug's MessageFactory interface to decouple myself from any PSR7 implementation.

src/Payum/Core/CoreGatewayFactory.php
Nyholm: Thank you.

composer.json
makasim: That would be great

Nyholm:
I've updated the PR according to your suggestions. We are now completely independent from Guzzle. Does it look good? Should I start writing some docs?

src/Payum/Be2Bill/Api.php
```diff
@@ -138,6 +144,7 @@ public function __construct(array $options, HttpClientInterface $client = null)
 
         $this->options = $options;
         $this->client = $client ?: HttpClientFactory::create();
+        $this->messageFactory = $options['httplug.message_factory'] ?: MessageFactoryDiscovery::find();;
```
makasim: I'd prefer to have it injected as third argument (optional) and if not passed use a discovery logic. FYI I am going to make this argument not optional in 2.x version as they managed by gateway factory.

src/Payum/Core/CoreGatewayFactory.php
```diff
@@ -59,10 +62,21 @@ public function createConfig(array $config = array())
...
+                return MessageFactoryDiscovery::find();
+            },
+            'payum.http_client'=>function(ArrayObject $config) {
+                $func = $config['httplug.client'];
+                if (is_callable($func)) {
+                    $func = $func($config);
+                }
 
-            'payum.http_client' => HttpClientFactory::create(),
-            'guzzle.client' => HttpClientFactory::createGuzzle(),
```
makasim: Could we add this back `guzzle.client` with a closure which throws an exception with a message why it does not exist any more.

makasim: and deprecated note.

makasim:
> Does it look good? Should I start writing some docs?

Pretty good I must say. There is one comment about the injection of message factory only.

src/Payum/Be2Bill/Api.php
Nyholm: Okey, will do.

src/Payum/Core/CoreGatewayFactory.php
Nyholm: Yes. I will

Nyholm:
I made the changes you suggested. Also, I added a HttplugClient and let the old HttpClient be. That means we do not break BC at all.

src/Payum/Be2Bill/Api.php
```diff
@@ -138,6 +144,7 @@ public function __construct(array $options, HttpClientInterface $client = null)
```
Nyholm: Oh. I forgot this line

Nyholm: ??  Sorry no. Just some weird bahavior with the phone

composer.json
```diff
@@ -39,7 +39,10 @@
     "require": {
         "php": "^5.5.0|^7.0",
         "payum/iso4217": "~1.0",
-        "guzzlehttp/guzzle": "~6.0",
+        "php-http/httplug": "^1.0",
```
makasim: any reason to require it explicitly. I personally dont think we need it since we have a `php-http/client-implementation` here. Do I miss something?

Nyholm: You are correct. This package is required indirectly. This row will not have any effect in practice. 

The reason I added it is because I believe it is good practice to specify all your dependencie and not rely on other packages to pull in the dependencies. We are using classes/interfaces from this package and that is why I've specified it as a dependency. 

I've noticed that this is very opinionated topic and I'm fine with removing it if you like.

makasim: let's keep it.

src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
```diff
@@ -1,26 +1,42 @@
...
+     * Create a Guzzle client.
+     *
+     * @return \GuzzleHttp\Client
      */
     public static function createGuzzle()
     {
+        trigger_error('The function "HttpClientFactory::createGuzzle" is depcrecated and will be removed in 2.0.', E_USER_DEPRECATED);
+
+        $client = null;
+        if (!class_exists('GuzzleHttp\Client')) {
```
makasim: Client::class

Nyholm: I can't. Guzzle might not be installed. If I'm using `Client::class` i would get an error saying "Class not found" instead of my custom exception.

src/Payum/Core/CoreGatewayFactory.php
```diff
@@ -59,10 +63,26 @@ public function createConfig(array $config = array())
...
+                return HttpClientDiscovery::find();
+            },
+            'httplug.message_factory'=>function(ArrayObject $config) {
+                return MessageFactoryDiscovery::find();
+            },
+            'payum.http_client'=>function(ArrayObject $config) {
+                $func = $config['httplug.client'];
+                if (is_callable($func)) {
+                    $func = $func($config);
+                }
```
makasim: @Nyholm you dont have to do it here since `httplug.client` will be resolved before the payum.http_client [here](https://github.com/Payum/Payum/pull/505/files#diff-d6a9354452023166c9054c19a425a2c8R134).

src/Payum/Core/PayumBuilder.php
```diff
@@ -373,6 +380,18 @@ public function setHttpClient(HttpClientInterface $httpClient = null)
...
     /**
+     * @param MessageFactory $httpMessageFactory
+     *
+     * @return PayumBuilder
+     */
+    public function setHttpMessageFactory(MessageFactory $httpMessageFactory)
+    {
+        $this->httpMessageFactory = $httpMessageFactory;
+
+        return $this;
```
makasim: No need for this I believe. If user want s to overwrite the default http client he can do it via `PayumBuilder:: addCoreGatewayFactoryConfig()` method. I am going to remove similar method `PayumBuilder::setHttpClient()` too.

src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
makasim: Hm, I think you can

src/Payum/Core/PayumBuilder.php
Nyholm: Okey, thank you

src/Payum/Core/CoreGatewayFactory.php
Nyholm: Without the check if it is a callable or not I got this error: https://travis-ci.org/Payum/Payum/jobs/117725270#L743

makasim: The test has to be fixed in that case. You should do a call_user_func there in the test.

Nyholm: Okey. I'll try. Thanks

src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
Nyholm: Yes, You are correct. I did not know that.

Nyholm:
Ive updated the docs, squashed my commits and also updated the PR description.

src/Payum/Core/Bridge/Guzzle/HttpClientFactory.php
```diff
@@ -28,6 +46,6 @@ public static function createGuzzle()
      */
     public static function create()
     {
-        return new HttpClient(static::createGuzzle());
+        return new HttplugClient(HttpClientDiscovery::find());
```
makasim: this change should be reverted. The whole factory is not used anywhere right now.

src/Payum/Core/CoreGatewayFactory.php
```diff
@@ -59,10 +66,37 @@ public function createConfig(array $config = array())
...
+                if (class_exists(MessageFactoryDiscovery::class)) {
+                    return MessageFactoryDiscovery::find();
+                }
+
+                throw new \LogicException('The httpplug.message_factory could not be guessed. Install one of the following packages: php-http/guzzle6-adapter. You can also overwrite the config option with your implementation.');
+            },
+            'payum.http_client'=>function(ArrayObject $config) {
+                  return new HttplugClient($config['httplug.client']);
+            },
             'guzzle.client' => HttpClientFactory::createGuzzle(),
```
makasim: Remove it from the config, with it we are sticked to guzzle, which is bad

src/Payum/Paypal/ExpressCheckout/Nvp/composer.json
```diff
@@ -24,7 +24,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
-        "symfony/phpunit-bridge": "~2.8|~3.0"
+        "symfony/phpunit-bridge": "~2.8|~3.0",
+        "guzzlehttp/psr7": "^1.1"
```
makasim: remove it.

src/Payum/Paypal/Ipn/composer.json
```diff
@@ -24,7 +24,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
-        "symfony/phpunit-bridge": "~2.8|~3.0"
+        "symfony/phpunit-bridge": "~2.8|~3.0",
+        "guzzlehttp/psr7": "^1.1"
```
makasim: remove it

src/Payum/Paypal/ExpressCheckout/Nvp/Api.php
```diff
@@ -303,10 +312,11 @@ class Api
     );
 
     /**
-     * @param array                $options
+     * @param array                    $options
      * @param HttpClientInterface|null $client
+     * @param MessageFactory|null      $messageFactory
      */
-    public function __construct(array $options, HttpClientInterface $client = null)
+    public function __construct(array $options, HttpClientInterface $client = null, MessageFactory $messageFactory = null)
```
Nyholm: Should this still be optional? If so, we need to require discovery in composer.json.

src/Payum/Core/CoreGatewayFactory.php
```diff
@@ -59,10 +66,37 @@ public function createConfig(array $config = array())
...
+                throw new \LogicException('The httpplug.client could not be guessed. Install one of the following packages: php-http/guzzle6-adapter. You can also overwrite the config option with your implementation.');
+            },
+            'httplug.message_factory'=>function(ArrayObject $config) {
+                if (class_exists(GuzzleClient::class)) {
+                    return new GuzzleMessageFactory();
+                }
+
+                // TODO add "if else" for other message factories provided by httplug.
+                
+                if (class_exists(MessageFactoryDiscovery::class)) {
```
makasim: It would be good to try discovery first and only after fallback to "my solution"

src/Payum/Paypal/ExpressCheckout/Nvp/Api.php
```diff
@@ -322,6 +332,7 @@ public function __construct(array $options, HttpClientInterface $client = null)
 
         $this->options = $options;
         $this->client = $client ?: HttpClientFactory::create();
+        $this->messageFactory = $messageFactory ?: MessageFactoryDiscovery::find();
```
Nyholm: ... We need to require discovery. Or, make our own Httplug factory with the logic from CoreGatewayFactory

src/Payum/Core/Resources/docs/get-it-started.md
```diff
@@ -13,13 +13,15 @@ The preferred way to install the library is using [composer](http://getcomposer.
 Run composer require to add dependencies to _composer.json_:
 
 ```bash
-php composer.phar require payum/offline
+php composer.phar require payum/offline php-http/guzzle6-adapter
```
makasim: `php-http/guzzle6-adapter` has to be added to all gateway's get-it-started docs.

src/Payum/Paypal/ExpressCheckout/Nvp/Api.php
makasim: Let's not use optional here. So no discovery

Nyholm:
There you go, I've added more Httplug clients.

src/Payum/Be2Bill/Api.php
```diff
@@ -120,10 +124,11 @@ class Api
     /**
      * @param array               $options
      * @param HttpClientInterface $client
+     * @param MessageFactory      $messageFactory
      *
      * @throws \Payum\Core\Exception\InvalidArgumentException if an option is invalid
      */
-    public function __construct(array $options, HttpClientInterface $client = null)
+    public function __construct(array $options, HttpClientInterface $client, MessageFactory $messageFactory )
```
Nyholm: Before I rewrite the failing tests. Is this the function signature that yo want?

makasim: yeap

Nyholm:
There. Everything we discussed is updated

UPGRADE.md
```diff
@@ -1,5 +1,9 @@
 # Upgrades
 
+## 1.3.0
+
+* [http-client] guzzlehttp/guzzle was decoupled and replaced by [Httplug](http://docs.php-http.org/en/latest/httplug/introduction.html). This gives you option to use Guzzle6, Guzzle5, Buzz or any other library to send HTTP messages. Make sure to specify `httplug.client` and `httplug.message_factory` or use [Httplug discovery](http://docs.php-http.org/en/latest/discovery.html). 
+
```
makasim: The text is not accurate for current state. I'd like this one:

```
* [http-client] When you update to Payum 1.3.0 the installation will fail because you need to install a client implementation. If you choose php-http/guzzle6-adapter everything will work exactly as before.
```

Nyholm: Updated

makasim:
@Nyholm Thanks for your work!

makasim:
That's a great move forward.

Nyholm:
Thank you for the feedback and for merging!

PR #506, opened 2016-03-22
Replace deprecated League\Url with League\Uri

skornev:
- Url null parameters  are used to be omitted now they are just empty
   Url lib:    ['var1'=>null]  -->  http://example.com/  
   Uri lib:    ['var1'=>null]  -->  http://example.com/?var1  
- Uri's are normalized out of the box. http://uri.thephpleague.com/uri/instantiation/
  http://example.com:80/xxx   -->   http://example.com/xxx

### DIFF
--- a/src/Payum/Be2Bill/Action/CaptureOffsiteNullAction.php
+++ b/src/Payum/Be2Bill/Action/CaptureOffsiteNullAction.php
@@ -1,7 +1,8 @@
 <?php
 namespace Payum\Be2Bill\Action;
 
-use League\Url\Url;
+use League\Uri\Schemes\Http as HttpUri;
+use League\Uri\Components\Query;
 use Payum\Core\Action\GatewayAwareAction;
 use Payum\Core\Exception\RequestNotSupportedException;
 use Payum\Core\Reply\HttpRedirect;
@@ -39,10 +40,11 @@ class CaptureOffsiteNullAction extends GatewayAwareAction
 
         $this->gateway->execute($getToken = new GetToken($extraData['capture_token']));
 
-        $url = Url::createFromUrl($getToken->getToken()->getTargetUrl());
-        $url->setQuery($httpRequest->query);
+        $uri = HttpUri::createFromString($getToken->getToken()->getTargetUrl());
+        $uri = $uri->withQuery(Query::createFromArray($httpRequest->query)->__toString());
 
-        throw new HttpRedirect((string) $url);
+
+        throw new HttpRedirect((string) $uri);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/Payum/Be2Bill/Action/CaptureOffsiteNullAction.php
```diff
@@ -39,10 +40,11 @@ public function execute($request)
 
         $this->gateway->execute($getToken = new GetToken($extraData['capture_token']));
 
-        $url = Url::createFromUrl($getToken->getToken()->getTargetUrl());
-        $url->setQuery($httpRequest->query);
+        $uri = HttpUri::createFromString($getToken->getToken()->getTargetUrl());
+        $uri = $uri->withQuery(Query::createFromArray($httpRequest->query)->__toString());
```
makasim: Looks better:

```php
<?php
$uri = $uri->withQuery((string) Query::createFromArray($httpRequest->query));
``

File: src/Payum/Core/Security/AbstractTokenFactory.php [hunk 2]
```diff
@@ -53,10 +54,12 @@ public function createToken($gatewayName, $model, $targetPath, array $targetPara
         }
 
         if ($afterPath && 0 === strpos($afterPath, 'http')) {
-            $afterUrl = Url::createFromUrl($afterPath);
-            $afterUrl->getQuery()->modify($afterParameters);
+            $afterUri = HttpUri::createFromString($afterPath);
+
+            $modifier = new MergeQuery(Query::createFromArray($afterParameters)->__toString());
+            $afterUri = $modifier->__invoke($afterUri);
```
makasim:
```php
<?php
$modifier($afterUri);
```

src/Payum/Be2Bill/Action/CaptureOffsiteNullAction.php
skornev: Your are right. I was thinking about that. They use __toString() in their examples thus I decided to keep it, but I did not change  (string) $uri.

src/Payum/Core/Security/AbstractTokenFactory.php
skornev: In this case I am no sure about that. For me it's less readable but I am not a big fan of magic methods.

makasim: I am also fine with something like this `call_user_func($modifier, $afterUri);`. Explicit call of magic method `__invoke` looks weird  to me.

makasim:
Thanks

PR #507, opened 2016-03-24
[Symfony Bridge] Action for Symfony Templating Component

davidfuhr:
Payum currently has a hard dependency on Twig. Adding this Action allows using other template engines easily. In the PayumBundle we could add a service definition like this...

```yml
services:
    payum.render_action:
        class: Payum\Core\Bridge\Symfony\Action\RenderTemplateAction
        arguments: ["@templating", '@PayumCore/layout.html.twig']
        tags:
            - { name: payum.action, all: true, alias: payum.action.render_template }
```

...and override the default twig render action.

This is backward compatible as long as twig is enabled in Symfony. And it also allows overriding the templates and using other template engines.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,6 +57,7 @@
         "symfony/http-kernel": "~2.8|~3.0",
         "symfony/http-foundation": "~2.8|~3.0",
         "symfony/form": "~2.8|~3.0",
+        "symfony/templating": "~2.8|~3.0",
         "symfony/validator": "~2.8|~3.0",
         "symfony/phpunit-bridge": "~2.8|~3.0",
         "symfony/dependency-injection": "~2.8|~3.0",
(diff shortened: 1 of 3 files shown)


makasim:
This is great! Shouldn't we overwrite it by default in the bundle? here: https://github.com/Payum/PayumBundle/blob/master/DependencyInjection/PayumExtension.php#L122


Could you add some tests?

davidfuhr:
I've pushed some tests.

If we override the render_template action in the PayumExtension the `payum.template.layout` variable becomes useless. To use different layout settings we would need different renderer instances... one per layout... where should this be implemented?

makasim:
We can use a builder for the action, like I did for obtain credit card one: https://github.com/Payum/Payum/blob/master/src/Payum/Core/Bridge/Symfony/Builder/ObtainCreditCardActionBuilder.php#L38

You can take a layout from the payum's config passed to the builder.

makasim:
Thanks.

PR #522, opened 2016-04-11
Add EventDispatcherExtension doc

davidfuhr:


makasim:
Would be good to add a section how to configure it in symfony app.

davidfuhr:
Shouldn't PayumBundle specific stuff go to PayumBundle documentation?

makasim:
This is about few lines of doc and could be placed here. If the doc grows we can add a dedicated chapter in the bundle. There is no any special rules, if you think it should be in the bundle, I am fine with it too.

makasim:
Thanks

PR #523, opened 2016-04-11
fixed usage of PENDINGREASON (paymentinfo instead of paymentrequest)

toooni:
@makasim As you can see here (https://developer.paypal.com/docs/classic/api/merchant/DoCapture_API_Operation_NVP/) PENDINGREASON is a Payment Info field, not a Payment Request field. Capturing an authorized payment does not work at the moment.


makasim:
I suspect this also fixes https://github.com/Payum/PayumBundle/issues/360

makasim:
Thanks

toooni:
@makasim do you know how long it takes until this change is on the subtree split? btw. why is the architecture like this? it's very complicated to contribute for me :( (maybe I just don't get it)

makasim:
I do split manually, usually while releasing a new version. run it now should be ok in few minutes.

toooni:
@makasim sorry to bother you, https://github.com/Payum/PaypalExpressCheckoutNvp is still not up to date.
Are you sure you want to keep it like that? It might be a bit discouraging to people wanting to contribute.

makasim:
done https://github.com/Payum/PaypalExpressCheckoutNvp/commits/master

makasim:
@toooni do you have an idea how it could be improved? 
Having stuff in one place makes my life a lot easier

makasim:
And this is like symfony does.

toooni:
@makasim hmm.. i just realized that it might be very hard to maintain all the payment "plugins" and it's versions for you.
Maybe the thing you could do is having a link on top of the subtree split repositories to an explanation how to create a PR from a branch of a subtree split.

makasim:
@toooni Available since 1.3.1

makasim:
@toooni https://github.com/Payum/Payum/pull/527. I havent tested it yet but something like this might works

toooni:
@makasim great 🎉  I'll test it soon..

PR #530, opened 2016-04-19
Keep already created gateway instances

### LINKED ISSUE #529: DynamicRegistry always creates new Gateway instances
The `DynamicRegistry` creates new Gateway instances every time `getGateway()` is called.

This is unexpected behaviour for a Registry (http://martinfowler.com/eaaCatalog/registry.html). And different from the behaviour of `AbstractRegistry`.

davidfuhr:
The expected behaviour of a registry is to keep state. Unless changed
from the outside. So we keep instances created earlier and return them
on request.

Fixes #529

### DIFF
--- a/src/Payum/Core/Registry/DynamicRegistry.php
+++ b/src/Payum/Core/Registry/DynamicRegistry.php
@@ -1,11 +1,17 @@
 <?php
 namespace Payum\Core\Registry;
 
+use Payum\Core\GatewayInterface;
 use Payum\Core\Model\GatewayConfigInterface;
 use Payum\Core\Storage\StorageInterface;
 
 class DynamicRegistry implements RegistryInterface
 {
+    /**
+     * @var GatewayInterface[]
+     */
+    private $gateways = [];
+
     /**
      * @var StorageInterface
      */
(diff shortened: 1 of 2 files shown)


src/Payum/Core/Registry/DynamicRegistry.php
```diff
@@ -64,7 +77,10 @@ public function getGateway($name)
      */
     public function getGateways()
     {
-        return $this->staticRegistry->getGateways();
+        return array_replace(
```
makasim: the behavior of the getGateways method depends on was getGateway method is called. Let's improve getGateways in a separate PR. https://github.com/Payum/Payum/issues/531

makasim: I am going to completely remove static registry from here, if you need both of them you have to use https://github.com/Payum/Payum/blob/master/src/Payum/Core/Registry/FallbackRegistry.php

The builder already works this way.

davidfuhr: Changes has been reverted.

makasim:
thank you

PR #540, opened 2016-04-30
Make storage extension to be state less.

### LINKED ISSUE #528: StorageExtension is not safe with multiple Gateways
There is only one `StorageExtension` instance per Model which is used in all Gateway instances. The `StorageExtension` keeps the Action Stack Level internally but this only works reliable when used with only one Gateway Instance. This leads to very unexpected behaviour.

Example

```php
<?php
//example.php

require __DIR__ . '/vendor/autoload.php';

use Payum\Core\PayumBuilder;
use Payum\Core\Payum;
use Payum\Core\Model\Payment;

$paymentClass = Payment::class;

/** @var Payum $payum */
$payum = (new PayumBuilder())
    ->addDefaultStorages()
    ->addGateway('aGateway', [
        'factory' => 'offline',
    ])
    ->addGateway('bGateway', [
        'factory' => 'offline',
    ])
    ->getPayum()
;

header('Content-Type: text/plain');

function getStorageExtension($gateway)
{
    $r = new ReflectionObject($gateway);
    $p = $r->getProperty('extensions');
    $p->setAccessible(true);
    foreach ($p->getValue($gateway) as $extension) {
        if ($extension instanceof \Payum\Core\Extension\StorageExtension) {
            return $extension;
        }
    }
}

$aGateway = $payum->getGateway('aGateway');
$aStorageExtension = getStorageExtension($aGateway);

$bGateway = $payum->getGateway('bGateway');
$bStorageExtension = getStorageExtension($bGateway);

var_dump($aStorageExtension === $bStorageExtension);
// true
```

makasim:
fixes https://github.com/Payum/Payum/issues/528

### DIFF
--- a/src/Payum/Core/Extension/Context.php
+++ b/src/Payum/Core/Extension/Context.php
@@ -36,6 +36,8 @@ class Context
      * @var Context[]
      */
     protected $previous;
+    
+    protected $values = [];
 
     /**
      * @param GatewayInterface $gateway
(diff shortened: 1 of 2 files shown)


makasim:
ping @davidfuhr

PR #542, opened 2016-05-04
Fixed default client creation issue for CurlAdapter

Hellsigner:
- Fixed default client creation issue for CurlAdapter;
 - Added httplug.stream_factory to default options

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,7 +70,7 @@
         "klarna/checkout": "~1|~2.0",
         "fp/klarna-invoice": "0.1.*",
         "stripe/stripe-php": "~2.0|~3.0",
-        "php-http/guzzle6-adapter": "^1.0",
+        "php-http/guzzle6-adapter": "^1.1",
         "ext-soap": "*"
     },
     "replace": {
--- a/composer.json
+++ b/composer.json
@@ -70,7 +70,7 @@
         "klarna/checkout": "~1|~2.0",
         "fp/klarna-invoice": "0.1.*",
         "stripe/stripe-php": "~2.0|~3.0",
-        "php-http/guzzle6-adapter": "^1.1",
+        "php-http/guzzle6-adapter": "^1.1.1",
         "ext-soap": "*"
     },
     "replace": {
--- a/src/Payum/Core/CoreGatewayFactory.php
+++ b/src/Payum/Core/CoreGatewayFactory.php
@@ -87,7 +87,7 @@ class CoreGatewayFactory implements GatewayFactoryInterface
                     return new DiactorosMessageFactory();
                 }
 
-                throw new \LogicException('The httplug.message_factory could not be guessed. Install one of the following packages: php-http/guzzle6-adapter. You can also overwrite the config option with your implementation.');
+                throw new \LogicException('The httplug.message_factory could not be guessed. Install one of the following packages: php-http/guzzle6-adapter, php-http/curl-client. You can also overwrite the config option with your implementation.');
             },
             'httplug.stream_factory'=>function(ArrayObject $config) {
                 if (class_exists(StreamFactoryDiscovery::class)) {
@@ -102,7 +102,7 @@ class CoreGatewayFactory implements GatewayFactoryInterface
                     return new DiactorosStreamFactory();
                 }
 
-                throw new \LogicException('The httplug.stream_factory could not be guessed. Install one of the following packages: php-http/guzzle6-adapter. You can also overwrite the config option with your implementation.');
+                throw new \LogicException('The httplug.stream_factory could not be guessed. Install one of the following packages: php-http/guzzle6-adapter, php-http/curl-client. You can also overwrite the config option with your implementation.');
             },
             'httplug.client'=>function(ArrayObject $config) {
                 if (class_exists(HttpClientDiscovery::class)) {
--- a/src/Payum/Core/CoreGatewayFactory.php
+++ b/src/Payum/Core/CoreGatewayFactory.php
@@ -110,7 +110,7 @@ class CoreGatewayFactory implements GatewayFactoryInterface
                 }
 
                 if (class_exists(HttpGuzzle6Client::class)) {
-                    return new HttpGuzzle6Client();
+                    return HttpGuzzle6Client::createWithConfig(array());
                 }
 
                 if (class_exists(HttpGuzzle5Client::class)) {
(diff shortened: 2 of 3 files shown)


src/Payum/Core/CoreGatewayFactory.php
```diff
@@ -89,7 +122,13 @@ public function createConfig(array $config = array())
                 }
 
                 if (class_exists(HttpCurlClient::class)) {
-                    return new HttpCurlClient();
+                    if ($config['httplug.message_factory'] instanceof \Closure) {
+                        $config['httplug.message_factory'] = $config['httplug.message_factory']->__invoke($config);
```
makasim: this not correct, the value must be already resolved.

makasim: make sure `httplug.stream_factory` and `httplug.message_factory` are added here prior http client option.

https://github.com/Payum/Payum/blob/master/src/Payum/Core/CoreGatewayFactory.php#L170

Hellsigner: > this not correct, the value must be already resolved.
XDebug shown that it's not, that's why I perform this check. It shall not work without it, because closures are passed here.

> make sure httplug.stream_factory and httplug.message_factory are added here prior http client option.
These options are placed in order of using in config dependencies, but when these closures are invoked, the order breaks by implicit reason, so I leaved initialization check explicit to make it work properly and sure.

makasim: read the second comment first.

Hellsigner: Sorry, skipped the link while reading first time. Done!

src/Payum/Core/CoreGatewayFactory.php
```diff
@@ -71,6 +74,36 @@ public function createConfig(array $config = array())
...
+
+                if (class_exists(\GuzzleHttp\Psr7\Request::class)) {
+                    return new GuzzleStreamFactory();
+                }
+
+                if (class_exists(\Zend\Diactoros\Request::class)) {
+                    return new DiactorosStreamFactory();
+                }
+
+                throw new \LogicException('The httplug.stream_factory could not be guessed. Install one of the following packages: php-http/guzzle6-adapter. You can also overwrite the config option with your implementation.');
```
makasim: correct messages here and above

makasim: add diactors package

makasim:
could you add some tests for these config options: httplug.stream_factory and httplug.message_factory

src/Payum/Core/CoreGatewayFactory.php
```diff
@@ -102,7 +102,7 @@ public function createConfig(array $config = array())
                     return new DiactorosStreamFactory();
                 }
 
-                throw new \LogicException('The httplug.stream_factory could not be guessed. Install one of the following packages: php-http/guzzle6-adapter. You can also overwrite the config option with your implementation.');
+                throw new \LogicException('The httplug.stream_factory could not be guessed. Install one of the following packages: php-http/guzzle6-adapter, php-http/curl-client. You can also overwrite the config option with your implementation.');
```
makasim: php-http/curl-client -> zendframework/zend-diactoros

Hellsigner:
I have made some new tests and a workaround for Guzzle6 client factory (there were a bug, see https://github.com/php-http/guzzle6-adapter/pull/32).
You can check it out.

src/Payum/Core/CoreGatewayFactory.php
```diff
@@ -71,13 +74,43 @@ public function createConfig(array $config = array())
...
+                throw new \LogicException('The httplug.stream_factory could not be guessed. Install one of the following packages: php-http/guzzle6-adapter, zendframework/zend-diactoros. You can also overwrite the config option with your implementation.');
+            },
             'httplug.client'=>function(ArrayObject $config) {
                 if (class_exists(HttpClientDiscovery::class)) {
                     return HttpClientDiscovery::find();
                 }
 
                 if (class_exists(HttpGuzzle6Client::class)) {
-                    return new HttpGuzzle6Client();
+                    return HttpGuzzle6Client::createWithConfig(array());
```
makasim: @Hellsigner  why did you do this change?

src/Payum/Core/CoreGatewayFactory.php
```diff
@@ -98,21 +131,6 @@ public function createConfig(array $config = array())
 
                 throw new \LogicException('The httplug.client could not be guessed. Install one of the following packages: php-http/guzzle6-adapter. You can also overwrite the config option with your implementation.');
             },
-            'httplug.message_factory'=>function(ArrayObject $config) {
```
makasim: please dont do ordering changes to prevent such noize diffs.

src/Payum/Core/Tests/CoreGatewayFactoryTest.php
```diff
@@ -72,6 +75,72 @@ public function shouldAlwaysAddHttpClientAsApi()
...
+        $this->assertInstanceOf(\Closure::class, $config['httplug.message_factory']);
+
+        $rc = new \ReflectionClass(CoreGatewayFactory::class);
+        $config = ArrayObject::ensureArrayObject($config);
+
+        $method = $rc->getMethod('buildClosures');
+
+        $rc->newInstance();
+        $method->setAccessible(true);
+        $method->invoke($factory, $config);
```
makasim: the code above is too tricky. It is not acceptable for tests.

Could you do required stuff explicitly:
```php
<?php
$this->assertInstanceOf(\Closure::class, $config['httplug.message_factory']);
call_user_func($config['httplug.message_factory']);

```

hope you get the idea?

src/Payum/Core/CoreGatewayFactory.php
Hellsigner: They already fixed this bug in 1.1.1, but Payum requires php-http/guzzle6-adapter ^1.0 in require-dev.
Without this change the tests are failing, so we need to update dependency or to leave this workaround

src/Payum/Core/Tests/CoreGatewayFactoryTest.php
Hellsigner: Done, all the tests are passing

src/Payum/Core/CoreGatewayFactory.php
Hellsigner: Ok, I'll mark it up

composer.json
```diff
@@ -70,7 +70,7 @@
         "klarna/checkout": "~1|~2.0",
         "fp/klarna-invoice": "0.1.*",
         "stripe/stripe-php": "~2.0|~3.0",
-        "php-http/guzzle6-adapter": "^1.0",
+        "php-http/guzzle6-adapter": "^1.1",
```
makasim: @Hellsigner  must be `^1.1.1`

PR #543, opened 2016-05-24
Fix stripe auto open

benlumley:
The stripe modal doesn't auto-open on load for me, possibly due to ordering of the block output? 

Using $(document).ready fixes it for me at least, and probably others - I'm seeing the issue whilst using sylius/sylius-standard.


makasim:
Thanks.

PR #544, opened 2016-05-30
Add property to class GetHttpRequest

ibogdan94:
Based on the code, where you use this class, you forgot to add the variable headers

### DIFF
--- a/src/Payum/Core/Request/GetHttpRequest.php
+++ b/src/Payum/Core/Request/GetHttpRequest.php
@@ -14,6 +14,11 @@ class GetHttpRequest
     public $request;
 
     /**
+     * @var array
+     */
+    public $headers;
+
+    /**
      * @var string
      */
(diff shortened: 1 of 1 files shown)


makasim:
what headers do you exactly need?

ibogdan94:
Maxim, Payum\Core\Bridge\Symfony\Action\GetHttpRequestAction.php
Function updateRequest takes parameters GetHttpRequest $request, Request $httpRequest
Than you use properties form class GetHttpRequest.php, but property headers doesn't described in it

makasim:
What header do you need but it missed right now?

ibogdan94:
I don't need a header. I thought it would be logical to add a property to the class that you used in your code, but forgot to add in GetHttpRequest.php

makasim:
I did not forget. It was done on purpose.

Issue #557, opened 2016-07-05
Payum "The EntityManager is closed"


makasim:
Better handling of cases when EntityManager is closed. From here https://github.com/Sylius/Sylius/issues/5362#issuecomment-228876839

PR #558, opened 2016-07-06
Fix EntityManager connection closed issue.

makasim:
fixes https://github.com/Payum/Payum/issues/557


makasim:
fixed https://github.com/Payum/Payum/commit/6825131c3fbd5b12c1a5dc2dcf0275f52225dbf4

Issue #568, opened 2016-08-12
Make sure skeleton package uses latest solutions


makasim:
I found that the base api aware class use deprecated logic.

PR #578, opened 2016-09-12
[skeleton] Remove deprecated features from skeleton.

makasim:


makasim:
fixes https://github.com/Payum/Payum/issues/568

PR #579, opened 2016-09-13
Adds DoVoid PayPal express NVP method as the gateway's Cancel action.

iisisrael:
Resubmitting the changes mistakenly requested on the read-only [PaypalExpressCheckoutNvp](https://github.com/Payum/PaypalExpressCheckoutNvp/pull/44) subtree split.

### DIFF
--- a/src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/DoVoidAction.php
+++ b/src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/DoVoidAction.php
@@ -3,26 +3,11 @@ namespace Payum\Paypal\ExpressCheckout\Nvp\Action\Api;
 
 use Payum\Core\Bridge\Spl\ArrayObject;
 use Payum\Core\Exception\RequestNotSupportedException;
-use Payum\Core\GatewayAwareInterface;
-use Payum\Core\GatewayInterface;
+use Payum\Core\Exception\LogicException;
 use Payum\Paypal\ExpressCheckout\Nvp\Request\Api\DoVoid;
-use Payum\Paypal\ExpressCheckout\Nvp\Request\Api\GetTransactionDetails;
 
-class DoVoidAction extends BaseApiAwareAction implements GatewayAwareInterface
+class DoVoidAction extends BaseApiAwareAction
 {
-    /**
-     * @var GatewayInterface
-     */
-    protected $gateway;
-
-    /**
-     * {@inheritdoc}
-     */
-    public function setGateway(GatewayInterface $gateway)
-    {
-        $this->gateway = $gateway;
-    }
-
     /**
      * {@inheritdoc}
      */
@@ -32,20 +17,14 @@ class DoVoidAction extends BaseApiAwareAction implements GatewayAwareInterface
         RequestNotSupportedException::assertSupports($this, $request);
 
         $model = ArrayObject::ensureArrayObject($request->getModel());
-        $paymentRequestN = $request->getPaymentRequestN();
 
-        $fields = new ArrayObject([]);
-        $parentTransactionId = $model['PAYMENTREQUEST_'.$paymentRequestN.'_PARENTTRANSACTIONID'];
-        $fields['AUTHORIZATIONID'] = $parentTransactionId
-            ? $parentTransactionId
-            : $model['PAYMENTREQUEST_'.$paymentRequestN.'_TRANSACTIONID']
-        ;
+        if (null === $model['TRANSACTIONID']) {
+            throw new LogicException('TRANSACTIONID must be set. Has user not authorized this transaction?');
+        }
 
-        $fields->validateNotEmpty(['AUTHORIZATIONID']);
-
-        $this->api->doVoid((array) $fields);
-
-        $this->gateway->execute(new GetTransactionDetails($model, $paymentRequestN));
+        $model->replace(
+            $this->api->doVoid((array) $model)
+        );
     }
 
     /**
--- a/src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/DoVoidAction.php
+++ b/src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/DoVoidAction.php
@@ -1,13 +1,27 @@
 <?php
 namespace Payum\Paypal\ExpressCheckout\Nvp\Action\Api;
 
+use Payum\Core\Action\ActionInterface;
+use Payum\Core\ApiAwareInterface;
+use Payum\Core\ApiAwareTrait;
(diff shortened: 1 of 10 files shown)


src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/DoVoidAction.php
```diff
@@ -0,0 +1,61 @@
...
+namespace Payum\Paypal\ExpressCheckout\Nvp\Action\Api;
+
+use Payum\Core\Bridge\Spl\ArrayObject;
+use Payum\Core\Exception\RequestNotSupportedException;
+use Payum\Core\GatewayAwareInterface;
+use Payum\Core\GatewayInterface;
+use Payum\Paypal\ExpressCheckout\Nvp\Request\Api\DoVoid;
+use Payum\Paypal\ExpressCheckout\Nvp\Request\Api\GetTransactionDetails;
+
+class DoVoidAction extends BaseApiAwareAction implements GatewayAwareInterface
```
makasim: This is better to do this way.

```php
<?php
class DoVoidAction implements ActionInterface, ApiAwareInterface, GatewayAwareInterface
{
   use ApiAwareTrait;
   use GatewayAwareTrait;
   
   public function __construct()
   {
       $this->apiClass = Api:class;
   }
}

File: src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/DoVoidAction.php [hunk 2]
```diff
@@ -0,0 +1,61 @@
...
+        $fields['AUTHORIZATIONID'] = $parentTransactionId
+            ? $parentTransactionId
+            : $model['PAYMENTREQUEST_'.$paymentRequestN.'_TRANSACTIONID']
+        ;
+
+        $fields->validateNotEmpty(['AUTHORIZATIONID']);
+
+        $this->api->doVoid((array) $fields);
+
+        $this->gateway->execute(new GetTransactionDetails($model, $paymentRequestN));
```
makasim: this should not be here, instead the CancelAction must do a Sync sub request

src/Payum/Paypal/ExpressCheckout/Nvp/Action/CancelAction.php
```diff
@@ -0,0 +1,43 @@
+<?php
+namespace Payum\Paypal\ExpressCheckout\Nvp\Action;
+
+use Payum\Core\Bridge\Spl\ArrayObject;
+use Payum\Core\Request\Cancel;
+use Payum\Core\Exception\RequestNotSupportedException;
+use Payum\Paypal\ExpressCheckout\Nvp\Api;
+use Payum\Paypal\ExpressCheckout\Nvp\Request\Api\DoVoid;
+
+class CancelAction extends PurchaseAction
```
makasim: why it extends PurchaseAction?

iisisrael:
Maybe so, but I duplicated the [DoCaptureAction](https://github.com/Payum/Payum/blob/master/src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/DoCaptureAction.php) for consistency.  If there are performance or implementation concerns, wouldn't that be better as a separate issue?

src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/DoVoidAction.php
iisisrael: Maybe so, but I duplicated the [DoCaptureAction](https://github.com/Payum/Payum/blob/master/src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/DoCaptureAction.php) for consistency.  If there are performance or implementation concerns, wouldn't that be better as a separate issue?

makasim:
> Maybe so, but I duplicated the DoCaptureAction for consistency. If there are performance or implementation concerns, wouldn't that be better as a separate issue?
yeah I know, I am in the process of migration to the new concept, 
in any case the BaseApiAction classes will be removed.

The trait also provide a better exception message when the support api was not found.

src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/DoVoidAction.php
makasim: > Maybe so, but I duplicated the DoCaptureAction for consistency. If there are performance or implementation concerns, wouldn't that be better as a separate issue?

Yeah I know, I am in the process of migration to the new concept, 
in any case the BaseApiAction classes will be removed.

The trait also provide a better exception message when the support api was not found.

src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/DoVoidAction.php
iisisrael: I think I see what you mean - fix pushed.

src/Payum/Paypal/ExpressCheckout/Nvp/Action/CancelAction.php
iisisrael: Lazy copy of CaptureAction?  Fixed to extend GatewayAwareAction and execute a Sync after executing a DoVoid on all payment requests that are Authroized.

src/Payum/Paypal/ExpressCheckout/Nvp/Api.php
```diff
@@ -529,6 +534,26 @@ public function doCapture(array $fields)
...
+     *
+     * @param array $fields
+     *
+     * @return array
+     */
+    public function doVoid(array $fields)
+    {
+        $fields['METHOD']  = 'DoVoid';
+        if (empty($fields['AUTHORIZATIONID'])) {
+            $fields['AUTHORIZATIONID'] = $fields['TRANSACTIONID'];
```
makasim: this is logic must be moved to action, it is not api related stuff

src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/DoVoidAction.php
makasim: @iisisrael Could you please use traits approach

iisisrael: Done, thanks for the code snippet.

src/Payum/Paypal/ExpressCheckout/Nvp/Api.php
iisisrael: Done.  I also noticed the action was missing tests, so added those.  I think I covered all the cases.

PR #536, opened 2016-04-28
[RFC] [Paypal] Add Pro Hosted Solution

thiphamyp:
Website Payments Pro Hosted Solution Integration  - NVP

* Introduction:
https://developer.paypal.com/webapps/developer/docs/classic/products/website-payments-pro-hosted-solution
* Integration
https://www.paypalobjects.com/webstatic/en_GB/developer/docs/pdf/hostedsolution_uk.pdf
https://www.paypalobjects.com/webstatic/en_FR/developer/docs/pdf/integralevolution.pdf

[Sandbox](https://github.com/makasim/PayumBundleSandbox/pull/71)

This is my very first gateway on Payum. 
Thanks for your comments and suggestions

PR #584, opened 2016-09-16
Paypal prohosted

makasim:


makasim:
contains work from https://github.com/Payum/Payum/pull/536

PR #595, opened 2016-10-18
Add Stripe Refund Support

bbwebconsult:

PR #596, opened 2016-10-19
Fix Stripe's CreateCustomerExtension and ConvertPaymentAction

bboure:
CreateCustomerExtension should accept credit card information as an array in order to create the customer:

```
[
    'number' => '',
    'exp_month' => '',
    'exp_year' => '',
    'cvc' => '',
]
```

This will allow the creation of a customer if the credit card was passed with its details instead of as a token.

Also, ConvertPaymentAction, when receiving a tokenized Credit card,
should set the token into `$details["card"]` instead of
`$details["customer"]` in order to let CreateCustomerExtension
create a Customer from a tokenized card.

Note:
I have no previous experience with Stripe but I think that `customer` and `card` have both been replaced by the `source` parameter. The source should accept anything that is tokenized (customer, credit card) or plain credit card information.
`card` and `customer` still seems to work but we might also consider refactoring this at some point.

### DIFF
--- a/src/Payum/Core/Action/CapturePaymentAction.php
+++ b/src/Payum/Core/Action/CapturePaymentAction.php
@@ -21,8 +21,8 @@ class CapturePaymentAction extends GatewayAwareAction
 
         /** @var $payment PaymentInterface */
         $payment = $request->getModel();
-
         $this->gateway->execute($status = new GetHumanStatus($payment));
+
         if ($status->isNew()) {
             $this->gateway->execute($convert = new Convert($payment, 'array', $request->getToken()));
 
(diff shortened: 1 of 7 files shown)


bbwebconsult:
Looks like I broke some tests.
I'll fix that and probably add some if you think this change makes sense

gpou:
@makasim @bbwebconsult Those changes will be included also on a bigger PR that we'll make with all the new features we're adding for Stripe.
So I guess it's better to close this one, right @bbwebconsult ?
See https://github.com/Payum/Payum/pull/595

bbwebconsult:
@gpou I would rather make several small PRs with each separate new feature we are implementing.
That way we could easily discuss each feature separately.

makasim:
@bboure @gpou what should we do with this PR?

PR #602, opened 2016-11-23
Restructure the project documentation

settermjd:
I'm keen to help out with improving the quality of the documentation, as it was an impediment to working with the library, as someone completely new to the project. Given that, I'm submitting this PR as a first step in helping to improve the docs. 

This PR:

- Restructures the documentation layout so that it is easier to find & easier to access.


makasim:
I am not sure this a good idea to mix php files with doc files. What we can do is to move all the docs to a completely separate repository.

Also it is not that easy to move files, links inside files have to be corrected too.

settermjd:
Thanks for the point about the links inside the docs. I'll have a look through the files and address that. I don't see the harm in keeping code and docs together. It would mean that people don't need to download a separate repository. However, having them separated would be cleaner. I'm happy to go with whichever path you choose.

makasim:
I'd prefer to have it in on place. payum/payum repository docs root folder.

makasim:
close in favor https://github.com/Payum/Payum/pull/621

PR #610, opened 2016-12-29
Add STATUS_REVERSED

Great-Antique:
Added for handling payment reverses (chargebacks) by handling `NotifyAction` executing

_P.S. Not sure about constant value of `GetBinaryStatus::STATUS_REVERSED`_

### DIFF
--- a/src/Payum/Core/Request/GetBinaryStatus.php
+++ b/src/Payum/Core/Request/GetBinaryStatus.php
@@ -15,6 +15,8 @@ class GetBinaryStatus extends BaseGetStatus
 
     const STATUS_PENDING = 1024; // 2^10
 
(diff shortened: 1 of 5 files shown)


src/Payum/Core/Request/GetBinaryStatus.php
```diff
@@ -140,6 +142,22 @@ public function isCanceled()
     /**
      * {@inheritDoc}
      */
+    public function markReversed()
```
makasim: We cannot add those methods here because of BC promise. 
Second I am not sure it has to be added at all. Let me read more about those stuff.

Great-Antique:
If you think that this solution is not suitable - close PR.

The minimum required by my case changes are in #611

makasim:
closed in favore  #611

PR #621, opened 2017-02-02
[docs] Put all docs in one place.

makasim:


makasim:
fixes https://github.com/Payum/PayumBundle/pull/407/files

PR #626, opened 2017-02-09
[Hotfix] catchReply on gateway execute method

welcoMattic:


makasim:
@welcoMattic Could you explain the reasons behind this change?

makasim:
Well catchReply set to true the reply will be returned. and that is how it works right now. There is a typo in doc then.

welcoMattic:
IMHO, the doc is ok, but the logic behaviour is not.

The name of argument is `$catchReply`, I understand that if it's set to `true` I can `catch` the `HttpResponse`. No?

Issue #634, opened 2017-03-02
Provide ability to encrypt gateway config info stored to database.


makasim:
source: https://github.com/Sylius/Sylius/issues/7582

makasim:
@pjedrzejewski I am thinking about such approach: Create a decorator storage (which implements StorageInterface and takes a real storage as argument). The storage encode\decode data. It also checks whether the model a storage configured for is GatewayConfigInterface. 

Also it would be good to add an option to payum bundle config which if set wrapps a gateway config storage with the decorator. 

Make sense?

makasim:
implemented in https://github.com/Payum/Payum/pull/641

PR #635, opened 2017-03-08
[Core] Create encrypting gateway configuration storage

Arminek:
Create storage which is able to decrypt gateway configuration.

Related with https://github.com/Payum/Payum/issues/634

### DIFF
--- a/src/Payum/Core/Storage/EncryptingGatewayConfigurationStorage.php
+++ b/src/Payum/Core/Storage/EncryptingGatewayConfigurationStorage.php
@@ -1,0 +1,134 @@
+<?php
+
+namespace Payum\Core\Storage;
+
+use Payum\Core\Model\GatewayConfigInterface;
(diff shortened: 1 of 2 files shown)


makasim:
would you able to add docs as well? If not I can manage that myself.

src/Payum/Core/Storage/EncryptingGatewayConfigurationStorage.php
```diff
@@ -0,0 +1,130 @@
+<?php
+
+namespace Payum\Core\Storage;
+
+use Payum\Core\Model\GatewayConfigInterface;
+
+final class EncryptingGatewayConfigurationStorage implements StorageInterface
```
makasim: EncryptedGatewayConfigStorage

Arminek:
Dead end. Closing in favor of #641

PR #641, opened 2017-03-22
[security] Add ability to crypt data stored to database

makasim:
This is an alternative approach to this PR https://github.com/Payum/Payum/pull/635

Here's some benefits with my approach:

* The algo is not limited to openssl. One can use another one by implementing Cypher interface.
* Rely on `defuse/php-encryption` library. 
* Not only gateway configs could be secured but payment data or any other model (model has to implement CryptedInterface)
* No need for doctrine listener and manipulations with change sets. 
* It would work with previously not encrypted data, it will be encrypted while using.

### DIFF
--- a/src/Payum/Core/Security/CypherInterface.php
+++ b/src/Payum/Core/Security/CypherInterface.php
@@ -1,0 +1,23 @@
+<?php
+namespace Payum\Core\Security;
+
+interface CypherInterface
+{
+    /**
+     * This method decrypts the passed value.
+     *
+     * @param string $value
+     *
+     * @return string
+     */
+    public function decrypt($value);
+
+    /**
+     * This method encrypts the passed value.
+     *
+     * Binary data may be base64-encoded.
+     *
+     * @param string $value
+     */
+    public function encrypt($value);
+}
--- a/src/Payum/Core/Security/CypherInterface.php
+++ b/src/Payum/Core/Security/CypherInterface.php
@@ -18,6 +18,8 @@ interface CypherInterface
      * Binary data may be base64-encoded.
      *
      * @param string $value
+     *
+     * @return string
      */
     public function encrypt($value);
 }
--- a/src/Payum/Core/Storage/CryptoStorageDecorator.php
+++ b/src/Payum/Core/Storage/CryptoStorageDecorator.php
@@ -1,0 +1,103 @@
+<?php
+
+namespace Payum\Core\Storage;
+
+use Payum\Core\Security\CryptedInterface;
+use Payum\Core\Security\CypherInterface;
+
+final class CryptoStorageDecorator implements StorageInterface
+{
+    /**
+     * @var StorageInterface
+     */
+    private $realStorage;
+
+    /**
+     * @var CypherInterface
+     */
+    private $crypto;
+
+    /**
+     * @param StorageInterface $realStorage
+     * @param CypherInterface $crypto
+     */
+    public function __construct(StorageInterface $realStorage, CypherInterface $crypto)
+    {
+        $this->realStorage = $realStorage;
+        $this->crypto = $crypto;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create()
+    {
+        return $this->realStorage->create();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function support($model)
+    {
+        return $this->realStorage->support($model);
+    }
+
+    /**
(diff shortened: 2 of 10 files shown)


makasim:
ping @pjedrzejewski @Arminek @pamil @inssein

Arminek:
Definitely, it is better approach. I felt that this encrypting/decrypting in storage was some kind of workaround.

src/Payum/Core/Security/CypherInterface.php
```diff
@@ -0,0 +1,23 @@
...
+     * @return string
+     */
+    public function decrypt($value);
+
+    /**
+     * This method encrypts the passed value.
+     *
+     * Binary data may be base64-encoded.
+     *
+     * @param string $value
```
pamil: Missing return docblock?

src/Payum/Core/Tests/Storage/CryptoStorageDecoratorTest.php
```diff
@@ -0,0 +1,234 @@
...
+    {
+        return $this->getMock(CypherInterface::class);
+    }
+}
+
+class CryptedModel implements CryptedInterface
+{
+    public function decrypt(CypherInterface $cypher)
+    {
+        $cypher->decrypt('theEncryptedVal');
```
pamil: Missing return

src/Payum/Core/Tests/Storage/CryptoStorageDecoratorTest.php
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function encrypt(CypherInterface $cypher)
+    {
+        $cypher->encrypt('theVal');
```
pamil: Missing return

src/Payum/Core/Tests/Storage/CryptoStorageDecoratorTest.php
```diff
@@ -0,0 +1,234 @@
...
+        ;
+        $cypherMock
+            ->expects($this->exactly(2))
+            ->method('decrypt')
+            ->with('theEncryptedVal');
+        ;
+
+        $storage = new CryptoStorageDecorator($decoratedStorage, $cypherMock);
+
+        $storage->findBy([]);
```
pamil: Shouldn't we assert that this method still returns a list of models?

src/Payum/Core/Tests/Storage/CryptoStorageDecoratorTest.php
```diff
@@ -0,0 +1,234 @@
...
+        ;
+        $cypherMock
+            ->expects($this->once())
+            ->method('decrypt')
+            ->with('theEncryptedVal');
+        ;
+
+        $storage = new CryptoStorageDecorator($decoratedStorage, $cypherMock);
+
+        $storage->find('anId');
```
pamil: Shouldn't we assert that this method still returns a model?

src/Payum/Core/Tests/Storage/CryptoStorageDecoratorTest.php
makasim: If you are talking about Crypted::decrypt and Crypted::encrypt methods. They do not return anything by design, just update an internal state of the model.

pamil: Oh, all right, `CypherInterface` and `CryptedInterface` have methods name the same and I just messed it up.

makasim: I felt that it could be that way. The names are easy to mess up.

pjedrzejewski:
Looks great! Thanks for working on this Maksim!

src/Payum/Core/Storage/CryptoStorageDecorator.php
```diff
@@ -0,0 +1,121 @@
...
+    {
+        return $this->decoratedStorage->identify($model);
+    }
+
+    /**
+     * @param object $model
+     */
+    private function assertCrypted($model)
+    {
+        if (false == $model instanceof  CryptedInterface) {
```
makasim: I personally prefer `false ==`. If someone else makes a PR I ignore such things (as it is allowed by PSR1\2). When I code myself I use `false ==`.

makasim:
Merged. Will be available since 1.4 release.

PR #642, opened 2017-03-24
[stripe] by providing the amount, stripe will validate CVC code

frechina:
Stripe does not validate CVC code in front.

This PR adds the amount as a param of the createToken function call since Stripe Checkout has built-in behavior that runs CVD checks if you've provided an amount.

This way stripe will return an error message like "Your card's security code is incorrect." and the form would print it

### DIFF
--- a/src/Payum/Stripe/Resources/views/Action/obtain_js_token.html.twig
+++ b/src/Payum/Stripe/Resources/views/Action/obtain_js_token.html.twig
@@ -59,7 +59,7 @@
                 // This identifies your website in the createToken call below
                 Stripe.setPublishableKey({{ publishable_key|json_encode|raw }});
 
-                Stripe.card.createToken($form, stripeResponseHandler);
+                Stripe.card.createToken($form, {{ model.amount }}, stripeResponseHandler);
 
                 // Prevent the form from submitting with the default action
                 return false;


makasim:
Is it be a BC break? 
Should we force this behavior by default?
What if someone wants to disable it? 

Is there any docs to read about this?

frechina:
> Is it be a BC break?
I don't think so

> Should we force this behavior by default?
Yes, I would

> What if someone wants to disable it?
disable what? if you don't check the CVC code, ChargeAction is going to fail.
if your card is incorrect (this is why CVC check fails) then you shouldn't have any stripeToken (which is the card as is)

> Is there any docs to read about this?
this is the only information I found about this
https://www.masteringmodernpayments.com/blog/how-to-check-cvc-and-address-with-stripe-js

makasim:
Then a question. Why it is not shown in their examples as default solution (or I missed something). There must be a reason.

makasim:
I'll check the doc

makasim:
```
This doesn't protect you against the card getting declined when you go to charge the card later. All it says is that this card can be charged right now.
This is completely undocumented, which means it could potentially break on you at any time.
```

I dont want to be it enabled by default. What about adding a doc on how to overwrite template and customize it?

makasim:
```
Hello Pete,
if you do this, on debit cards the customers probably will se a charge ( which will expire after X days ) even if it isn't a charge and is only an auth.
That can cause some customers to freak out ...

Thanks,

```

frechina:
I'm not sure about this. I checked on Stripe Logs and I didn't see any 'extra' charge. I will research more about it

PR #664, opened 2017-06-15
Fix PayPal express checkout NVP Cancel action support scope

### LINKED ISSUE #660: Cancel recurring paypal express
Having a problem when i try to cancel a recurring paypal payment. 

When I do ```$gateway->execute(new Cancel($recurringPayment));``` nothing happens. I tried tracing the root of the problem and in ```ExpressCheckout/Nvp/Action\CancelAction.php``` there are these lines of code:

```
if (!$details['TRANSACTIONID']) {
    return;
}
```

A recurring payment will never have a transaction ID. So how is this supposed to work? Are the [docs](https://github.com/Payum/Payum/blob/master/docs/paypal/express-checkout/cancel-recurring-payment.md) wrong?

dincho:
Because of very wide scope of support method, this action was preventing execution of other actions that handle Cancel request like CancelRecurringPaymentsProfileAction

fix #660

### DIFF
--- a/src/Payum/Paypal/ExpressCheckout/Nvp/Action/CancelAction.php
+++ b/src/Payum/Paypal/ExpressCheckout/Nvp/Action/CancelAction.php
@@ -24,10 +24,6 @@ class CancelAction implements ActionInterface, GatewayAwareInterface
 
         $details = ArrayObject::ensureArrayObject($request->getModel());
 
-        if (!$details['TRANSACTIONID']) {
-            return;
-        }
-
         $voidDetails = new ArrayObject([
             'AUTHORIZATIONID' => $details['TRANSACTIONID'],
         ]);
@@ -41,9 +37,12 @@ class CancelAction implements ActionInterface, GatewayAwareInterface
      */
     public function supports($request)
     {
-        return
-            $request instanceof Cancel &&
-            $request->getModel() instanceof \ArrayAccess
-        ;
+        if (false == ($request instanceof Cancel && $request->getModel() instanceof \ArrayAccess)) {
+            return false;
+        }
+
+        // it must take into account only common payments, recurring payments must be cancelled by another action.
+        $model = $request->getModel();
+        return isset($model['TRANSACTIONID']);
     }
 }
--- a/src/Payum/Paypal/ExpressCheckout/Nvp/Action/CancelAction.php
+++ b/src/Payum/Paypal/ExpressCheckout/Nvp/Action/CancelAction.php
@@ -24,6 +24,10 @@ class CancelAction implements ActionInterface, GatewayAwareInterface
 
         $details = ArrayObject::ensureArrayObject($request->getModel());
 
+        if (!$details['TRANSACTIONID']) {
+            return;
+        }
+
         $voidDetails = new ArrayObject([
             'AUTHORIZATIONID' => $details['TRANSACTIONID'],
         ]);
@@ -43,6 +47,6 @@ class CancelAction implements ActionInterface, GatewayAwareInterface
 
         // it must take into account only common payments, recurring payments must be cancelled by another action.
         $model = $request->getModel();
-        return isset($model['TRANSACTIONID']);
+        return empty($model['BILLINGPERIOD']);
     }
 }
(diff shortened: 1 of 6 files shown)


src/Payum/Paypal/ExpressCheckout/Nvp/Action/CancelAction.php
```diff
@@ -41,9 +37,12 @@ public function execute($request)
...
-            $request instanceof Cancel &&
-            $request->getModel() instanceof \ArrayAccess
-        ;
+        if (false == ($request instanceof Cancel && $request->getModel() instanceof \ArrayAccess)) {
+            return false;
+        }
+
+        // it must take into account only common payments, recurring payments must be cancelled by another action.
+        $model = $request->getModel();
+        return isset($model['TRANSACTIONID']);
```
makasim: This is a BC break. before if you execute cancel request with the model that does not have a TRANSACTIONID you'll simply do nothing. Now it leads to not supported action exception. 

I'd suggest putting TRANSACTIONID back where it was and add a check `BILLINGPERIOD` is empty to supports method. 

btw could you please move this class out of Api folder. to a parent one while you are on it https://github.com/Payum/Payum/blob/master/src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/CancelRecurringPaymentsProfileAction.php

dincho: I agree it's BC break, but I was wondering why would you send cancel request with empty model. Will fix.

However, isn't it better to check for empty PROFILEID ?

dincho: > btw could you please move this class out of Api folder. to a parent one while you are on it https://github.com/Payum/Payum/blob/master/src/Payum/Paypal/ExpressCheckout/Nvp/Action/Api/CancelRecurringPaymentsProfileAction.php

I was hoping to discuss to re-factor this action a little bit, currently it does a replace on the model but the [ManageRecurringPaymentsProfileStatus response](https://developer.paypal.com/docs/classic/api/merchant/ManageRecurringPaymentsProfileStatus_API_Operation_NVP/#managerecurringpaymentsprofilestatus-response-fields) returns just the PROFILEID. It does not makes any sense since the PROFILEID won't change, however all the previous data in the model will be lost.

I think that better approach is to implement very similar 2 gateway calls like in [transaction Cancel action](https://github.com/Payum/Payum/blob/master/src/Payum/Paypal/ExpressCheckout/Nvp/Action/CancelAction.php#L35), what do you think?

makasim: > I think that better approach is to implement very similar 2 gateway calls like in transaction Cancel action, what do you think?

it makes sense to me

makasim: > However, isn't it better to check for empty PROFILEID ?

We can do it as long as a recurring agreement does not have it. So we can tell the difference.

dincho: Also moving the action namespace is BC break, do you have any convention for deprecation?

I could keep the old class extending the new class and add:
```php
trigger_error('The ' . __CLASS__ . ' class is deprecated, use . NewCancelAction::class . , it will be removed in the next major version.', E_USER_DEPRECATED);
```

makasim: > do you have any convention for deprecation?

Nothing yet but we could simply follow Symfony's convention.

dincho: > Nothing yet but we could simply follow Symfony's convention.

changed namespace in 509e1fe6, let me know if it works for you

dincho:
> I think that better approach is to implement very similar 2 gateway calls like in transaction Cancel action, what do you think?
> it makes sense to me

see 49f4d3f8

dincho:
@makasim anything else I should do to get this merged?

makasim:
Should be available since 1.4.2

PR #672, opened 2017-08-05
[Sofort] Fix status action and cover with tests, make docs correction

### LINKED ISSUE #670: [Sofort] Pending status marks the payment as captured
Maybe I'm missing something but it's confusing. In the `StatusAction` for Sofort integration it sets the payment as captured (instead of pending) when the Sofort status is `pending`. Should it be that way? If so why is that?

https://github.com/Payum/Payum/blob/master/src/Payum/Sofort/Action/StatusAction.php

```php
// ...
switch ($details['status']) {
            case Api::STATUS_LOSS:
                $request->markFailed();
                break;
            case Api::STATUS_PENDING:
                $request->markCaptured();
                break;
// ...
```

Tykzz:
Fixes #670

### DIFF
--- a/docs/sofort/disable-notifications.md
+++ b/docs/sofort/disable-notifications.md
@@ -13,8 +13,8 @@ When you are working locally with Sofort you may get
 </errors>
 ```
 
-That's because the notification url you sent to Sofort is not reachable and Sofort returns error in this case.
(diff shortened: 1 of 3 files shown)


makasim:
Thanks.

Issue #675, opened 2017-08-21
Update OmnipayBridge to Support Symfony 3.*


dpfaffenbauer:
Hi,

OmnipayBridge currently doesn't support Symfony 3 applications. The reason is the dependency to omnipay/common:^2.3 which has a dependency to symfony/http-foundation: ~2.1 and therefore conflicts with Symfony 3.*

Omnipay/Common already released a Alpha Version where they use symfony/http-foundation:^2.1|^3.

If you could do the same with the OmnipayBridge, it would work again.

Cheers

makasim:
Easier said than done. It would require extensive testing and changes checking before we can raise the supported version.

makasim:
There might be some BC breaks as well.

dpfaffenbauer:
Thats right.

Regardless BC breaks, Omnipay drivers have a dependency on omnipay/common:~2 as well. For example Worldpay: https://github.com/thephpleague/omnipay-worldpay/blob/master/composer.json

So, this is much more far-reaching than first thought.

makasim:
I am not up to the task cuz I do not work with omnipay at the moment. Though if someone willing to contribute I do my best to review it and get it merged.

dpfaffenbauer:
I have to correct the issue description: Problem appear cause of older guzzle version.

I am working on a fix, but as already said: this is gonna be BC break, therefore I would consider this as OmnipayBridge 2?

Here is what I came up with.

#### OmnipayBridge
Integrates new omnipay/common lib **BC break here**  (in OmnipayGatewayFactory: removed first argument, cause Omnipay doesn't tell us which gateways are available anymore, you'll find this in PayumCore (PayumBuilder) as well)
https://github.com/dpfaffenbauer/OmnipayBridge/commit/8e7c59912de73b19a70c7f913756b9995363294a
This removes deprecated classes (since this is a BC anyway, I thought we could remove them as well): https://github.com/dpfaffenbauer/OmnipayBridge/commit/4a9b160957adde6cd450697f5ff8f3e3c9b53afa

#### Payum Core
[...]

PR #679, opened 2017-09-01
Make Payum compatible with OmnipayBridge 2.0

dpfaffenbauer:
This PR makes Payum compatible with not yet released OmnipayBridge 2.0 (https://github.com/Payum/OmnipayBridge/pull/32).

It looks like a BC break, but since it affects only OmnipayBridge, I wouldn't consider it a BC break at all.

related to https://github.com/Payum/Payum/issues/675

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,8 +49,8 @@
     "require-dev": {
         "ext-curl": "*",
         "ext-pdo_sqlite": "*",
-        "payum/omnipay-bridge": "^1",
-        "omnipay/dummy": "~2.0",
+        "payum/omnipay-bridge": "^2.0",
+        "omnipay/dummy": "~3.0",
         "doctrine/orm": "2.5.*",
         "phpunit/phpunit": "~4.0",
         "propel/propel1": "~1.7",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 2.0.0 (2017-08-21)
+* [bridge][omnipay] make it compatible with ominpay/common:^3.0, since Omnipay removed getSupportedGateways() function, you now need add 'type' property to your gateway-config or pass it to the gateway factory.
+
 ## 1.4.2 (2017-06-21)
 
 * [paypal-ec] Re-factor CancelRecurringPaymentsProfileAction as actions proxy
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -49,8 +49,8 @@
     "require-dev": {
         "ext-curl": "*",
         "ext-pdo_sqlite": "*",
-        "payum/omnipay-bridge": "^1",
-        "omnipay/dummy": "~2.0",
+        "payum/omnipay-bridge": "^2.0",
+        "omnipay/dummy": "~3.0",
```
dpfaffenbauer: Not sure but: since omnipay/dummy and payum/omnipay-bridge (not yet) are still in Alpha, we would need to lower the minimum-stability here as well, wouldn't we?

makasim:
Wouldn't it better to create a new repo for omnipay v2 bridge?

dpfaffenbauer:
Well yeah, would make sense for the Bridge, but since PayumBuilder is kinda coupled to the Bridge, we need to make changes inside Payum Core as well.

I mean, PayumBuilder could support old and new version, we just need add another check here for OmnipayGateway2Factory 
 https://github.com/dpfaffenbauer/Payum/blob/bc766db23b3ddc7c919648436c37e36bb489c5d2/src/Payum/Core/PayumBuilder.php#L608.

makasim:
I'd suggest going this way:

1. create a new repo `OmnipayV2Bridge`
2. push the current bridge code into it
3. change namespace to `Payum\OmnipayV2Bridge;` so they do not collide. 
4. Do required changes.
5. Add support of the new bridge to the payum builder and other changes in the core (like adding it to composer suggests section)

dpfaffenbauer:
Alright, can you create the repo and push the current version into it? I can then create the PR for Bridge2 and PayumCore

makasim:
@dpfaffenbauer done: https://github.com/Payum/OmnipayV3Bridge

dpfaffenbauer:
Alright, PR for Bridge is out, only thing that could, or will, cause a problem now is: version constraints. since payum/payum includes the bridge in require-dev, we can't do that with both versions now, this will cause a version-constraint problem. what should we do about that?

makasim:
We can leave that part as is for now. when omnipay 2.x is out we can replace 1.x to 2.x in require-dev.

PR #681, opened 2017-09-11
Missing variables

tjurczak:
I noticed in documentation that in capture.php is missing initializate two variables which are used bellow.


makasim:
Thanks.

PR #695, opened 2017-10-19
Some docs improvements

ThomasLandauer:
Open question (please answer, if you can):

* "and delegate the job to capture action": Where is this action? And which "job" do you mean? The request?

### DIFF
--- a/docs/symfony/get-it-started.md
+++ b/docs/symfony/get-it-started.md
@@ -106,7 +106,7 @@ _**Note**: You can add other gateways to the gateways section too._
 
 ## Prepare order
 
-Now we can create an order. Create a capture token and delegate the job to capture action.
+Now we can create an order. In the last line the user is redirected to an URL which is handled by [`CaptureController ::doAction()`](https://github.com/Payum/PayumBundle/blob/fd930cb9516c8a5f19b4eeae35c8e37eea77ce11/Controller/CaptureController.php#L30)
 
 ```php
 <?php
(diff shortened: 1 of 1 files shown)


makasim:
First, thank you so much for your time and the improvements. The doc desperately needs polishing.

Regarding the questions. In this particular case by action, I meant an HTTP redirect (or Symfony's forward) to [capture action](https://github.com/Payum/PayumBundle/blob/master/Controller/CaptureController.php#L10). The action is in terms of Symfony (something that handles an HTTP request). 

> And which "job" do you mean

I meant that we had finished with the preparation and it was time to hand this over to capture logic.

Let me know if you have any other questions or my answers are not clear.

ThomasLandauer:
Um, there is no `captureAction` on https://github.com/Payum/Payum/blob/master/docs/symfony/get-it-started.md  - that's what puzzled me in the first place.  
And your link leads to `doSessionTokenAction()`

makasim:
@ThomasLandauer sorry The link must've been https://github.com/Payum/PayumBundle/blob/fd930cb9516c8a5f19b4eeae35c8e37eea77ce11/Controller/CaptureController.php#L30. My fault. 

> Create a capture token and delegate the job to capture action.

is mostly describes this line of code (the last line of the prepareAction()):

```php
<?php 

return $this->redirect($captureToken->getTargetUrl());  
```

ThomasLandauer:
I deleted the term "capture action" - it's not that important ;-)  
If the rest is correct, then please merge it, I'm done for now.

I'll eventually come up with more PR's on my way through the library :-)

makasim:
Thank you

PR #697, opened 2017-10-21
Get the sample code running

ThomasLandauer:


ThomasLandauer:
Another suggesion for the docs: Integrate "tabs" above the code-blocks, to switch between YAML, XML and annotations. Example: https://symfony.com/doc/current/reference/configuration/doctrine.html#custom-mapping-entities-in-a-bundle

docs/symfony/get-it-started.md
```diff
@@ -135,7 +166,7 @@ class PaymentController extends Controller
         $captureToken = $this->get('payum')->getTokenFactory()->createCaptureToken(
             $gatewayName, 
             $payment, 
-            'done' // the route to redirect after capture
```
makasim: nope it should be `done`.

docs/symfony/get-it-started.md
```diff
@@ -84,7 +84,36 @@ class Payment extends BasePayment
...
+
+Acme\PaymentBundle\Entity\Payment:
+    type: entity
+    table: payment
+
+    id:
+        id:
+            type: integer
+            generator:
+                strategy: IDENTITY
```
makasim: there are annotations on the model that do the same thing.

docs/symfony/get-it-started.md
ThomasLandauer: Hm, this is what my `\vendor\payum\payum-bundle\Resources\config\routing\capture.xml` looks like (Symfony):

```xml
<?xml version="1.0" encoding="UTF-8" ?>

<routes xmlns="http://symfony.com/schema/routing" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/routing http://symfony.com/schema/routing/routing-1.0.xsd">
    <route id="payum_capture_do_session" path="/payment/capture/session-token">
        <default key="_controller">PayumBundle:Capture:doSessionToken</default>
    </route>

    <route id="payum_capture_do" path="/payment/capture/{payum_token}">
        <default key="_controller">PayumBundle:Capture:do</default>
    </route>
</routes>
```

I have no route `done` there. And I got an error when I tried it with `done`, so I figured I have to change it to `payum_capture_do`

docs/symfony/get-it-started.md
ThomasLandauer: Yeah, sure. But if you have any YAML files in your bundle, annotations won't work anymore, see https://stackoverflow.com/a/26141187/1668200

That's why I suggested to show YML, XML, and Annotations in parallel.

docs/symfony/get-it-started.md
makasim: `done` route is one created by you in your app. It is not part of payum bundle.

makasim:
> Another suggesion for the docs: Integrate "tabs" above the code-blocks, to switch between YAML, XML and annotations. Example: 

That would be great but as far as I know, we cannot do it on Github in markdown.

ThomasLandauer:
Tabs: The Symfony docs (see link above) are done in GitHub! I don't have a place where I could try it out, so it's probably better if you do it ;-)

docs/symfony/get-it-started.md
ThomasLandauer: Uh, OK! Then this would be the right place to explain that :-)

* What is the route/controller supposed to do? (In which situation is it called?)
* Any parameters?

makasim:
The tabs are nice looking on the symfony.com site but on the github: https://github.com/symfony/symfony-docs/blob/master/reference/configuration/doctrine.rst

makasim:
We can create separate docs for xml, yaml configurations and link them from get it started doc.

ThomasLandauer:
OK, sorry, now I understand: GitHub can't *render* it nicely :-(  
Well, I wouldn't go for separate files, cause they will (sooner or later) run out of sync. Having 2-3 alternative code blocks underneath each other isn't that big a problem, I would say. Is Symfony the only framework with this issue?

For now: Let's get the `done` route question straight, so you can merge this PR. Maybe I have a better idea for the 3 configs when I see it "live".

docs/symfony/get-it-started.md
makasim: > What is the route/controller supposed to do? (In which situation is it called?)

The payment process should always end up there, regardless of payment status. Inside it, you have to check the payment status [like this] and do whatever your business dictates you to do. 
 (https://github.com/makasim/PayumBundleSandbox/blob/23200c1aab156eef0f7c1e3f2373ab7414dce27e/src/Acme/PaymentBundle/Controller/DetailsController.php#L27).

makasim: > Any parameters?

Nope, everything you need is bound to the token already  (a payment and its details)

ThomasLandauer: So the `done` route is where the user is being redirected to, when returning from the (e.g.) PayPal website? So that's what most payment providers call the "success url"?

ThomasLandauer: Sorry, what you're saying can't be right ;-) The purpose of this `prepareAction` is to set up the payment ("capture"), and then redirect the user to the payment provider, right? If I enter one of my own routes as this argument, the user gets redirected there. This doesn't make any sense, or does it?

makasim: > So the done route is where the user is being redirected to, when returning from the (e.g.) PayPal website? So that's what most payment providers call the "success url"?

nope, the user comes from Paypal back to capture url and only after it is redirect to done url (by capture action)

makasim: here a user is being redirected to done url https://github.com/Payum/PayumBundle/blob/master/Controller/CaptureController.php#L39

Issue #696, opened 2017-10-21
Remove suggested installations from composer


ThomasLandauer:
Please remove all those suggestions upon `composer require` like this one:

> payum/core suggests installing payum/paypal-rest

It doesn't make sense to install *everything*, you only need the payment packages for the payment methos you actually have!

makasim:
Okay, I don't mind.

makasim:
done in https://github.com/Payum/Payum/pull/701

PR #701, opened 2017-10-21
Remove `suggests`

ThomasLandauer:
See https://github.com/Payum/Payum/issues/696  
Please double-check; I've never done this before. I removed the entire `suggests` key.


makasim:
Looks good.

PR #702, opened 2017-10-23
Create overview

ThomasLandauer:


makasim:
It is only valid for Paypal-like payment providers, but there are other payment providers that work differently, for example stripe.js, klarna checkout, be2bill, payex

Issue #693, opened 2017-10-07
`idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated`


stefandoorn:
While testing Sylius against PHP 7.2 (https://github.com/Sylius/Sylius/pull/8771), I noticed a package gave a deprecation warning:

```
idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated in /../../vendor/league/uri/src/Components/HostnameTrait.php line 68
```

Payum is requiring `league/uri` at `~4.0`. It seems to me that this deprecation has been fixed on their side, but it requires upgrading to `~5.0`:

* https://github.com/thephpleague/uri/releases
* https://github.com/thephpleague/uri-parser/blob/82e0e63c3e3ee925d2cb92199443ec2805e7e564/CHANGELOG.md#fixed-4

As 4.x -> 5.x comes with BC breaks, this might not be an easy thing to fix. As I don't know about the internals of Payum well enough, I cannot provide a PR at this moment.

stefandoorn:
PHP 7.2 Changelog mention: http://php.net/manual/en/function.idn-to-ascii.php

pamil:
jeremykendall/php-domain-parser#204 and thephpleague/uri#105 are trying to solve this issue in the current Payum releases, however moving Payum to the most recent PHP versions would also help with issues like these one.

stefandoorn:
https://github.com/thephpleague/uri/issues/105 got merged and 4.2.3 got released. This version fixes the issue for PHP 7.2 and allows Payum also to be PHP 7.2 compatible imo.

PR #718, opened 2017-11-26
allow league/uri 5

garak:
This should solve #693


m-vo:
unfortunately there are breaking changes going from v4 to 5 / see #723 (which should hopefully work around them)

garak:
@m-vo thanks for letting me know. I close this one then

Issue #712, opened 2017-11-17
Support Symfony 4


garak:
See http://symfony.com/blog/helping-prepare-for-symfony-4-bundle-support
I tried a basic install and didn't find anythinig deprecated (but I didn't dig too deep).
Things to do:
* update composer.json
* test on travis
* explicit public/private services

ThomasLandauer:
Here's a deprecation warning I'm getting in Symfony 3.4:

> User Deprecated: Implementing "Symfony\Component\HttpKernel\DataCollector\DataCollectorInterface" without the "reset()" method is deprecated since version 3.4 and will be unsupported in 4.0 for class "Payum\Bundle\PayumBundle\Profiler\PayumCollector".

PR #726, opened 2017-12-22
Drop PHP5, Symfony 2. Add Symfony 4. Upgrade phpunit

makasim:


makasim:
partially fixes  https://github.com/Payum/Payum/issues/712

PR #735, opened 2018-02-06
[Paypal Express Checkout nvp] Introduce the TransactionSearch paypal api endpoint

BenoitLeveque:
we need to call this endpoint to know the payment status of a legacy recurring payment.

i've tried to use the `GetRecurringPaymentsProfileDetailsAction` action, but my recurring payment has been created from a legacy version of paypal.

here is the paypal documentation that explain how this endpoint work: https://developer.paypal.com/docs/classic/api/merchant/TransactionSearch_API_Operation_NVP/?mark=TransactionSearch


makasim:
LGTM
You should add it to the factory too https://github.com/Payum/Payum/blob/master/src/Payum/Paypal/ExpressCheckout/Nvp/PaypalExpressCheckoutGatewayFactory.php#L57

BenoitLeveque:
@makasim ok, i've added the missing line in the gateway factory

makasim:
Thank you.

PR #766, opened 2018-06-15
Bug fix - validation of notification url on Sofort

violuke:
Currently Payum provides a way to switch off notification URLs for Sofort (to aid development, or when it is not needed), however the build in validation within Payum then fails. This fixes that so you can use the disable notification option.


violuke:
Could this please be merged in? Thank you.

Issue #771, opened 2018-07-24
AnnotationException because of @Use in stripe action classes


TamasSzigeti:
Hi,

Thanking for the great lib!

I'm using payum/stripe with Symfony, and extended a couple action classes. The @Use in Stripe/Action/Api/*Action.php triggers this exception below in certain cases. I can't even suppress it, because @IgnoreAnnotation doesn't work for parent classes. It might be a good idea to use the @deprecated annotation instead.

In AnnotationException.php line 54:
[Semantical Error] The annotation "@Use" in property Payum\Stripe\Action\Api\ObtainTokenAction::$keys was never imported. Did you maybe forget to add a "use" statement for this annotation?

makasim:
Hi!

Feel free to submit a PR

PR #772, opened 2018-07-25
Fix deprecation comments, use proper phpdoc #771

TamasSzigeti:
Fixing #771


TamasSzigeti:
Thanking muchly for the quick merge!

Issue #760, opened 2018-05-10
HttpRequestVerifier fails when URL is prefixed with locale via JMSI18nRoutingBundle


TamasSzigeti:
https://github.com/Payum/Payum/blob/5df1fbf9ddfb0fa6c1879d4bdeec483c7640cf73/src/Payum/Core/Bridge/Symfony/Security/HttpRequestVerifier.php#L52

This check of comparing current URL to token URL fails when JMSI18nRoutingBundle locale prefixing is in use. I believe `$httpRequest->getUri()` should be replaced with `$httpRequest->getRequestUri()`, but I'm not confident enough that it wouldn't break anything so I'm creating an issue instead of a PR. Please check it out…

Thank you

TamasSzigeti:
Can someone look at this, please?

TamasSzigeti:
This issue has been stale for a while, I would gladly create a PR here as well, if you have no concerns about this change.

makasim:
@TamasSzigeti that would be great!

TamasSzigeti:
I was wrong, the error was caused by something else

makasim:
@TamasSzigeti have you found what caused the error?

TamasSzigeti:
Not yet, but I've put together a sandbox app with JMSI18nRoutingBundle and it works correctly, getUri matches getRequestUri. Will let you know when I figured out anything meaningful

PR #773, opened 2018-07-25
#760 Fix verifying URI

TamasSzigeti:
Fixing #760 – when JMSI18nRoutingBundle is prefixing URIs


makasim:
@TamasSzigeti could you please write a test that covers the issue?

TamasSzigeti:
A really good call, in my attempt to write a test for it I figured out that the problem was coming from somewhere else. Apologies for the clutter

Issue #797, opened 2019-03-03
Implement security checker in composer


liamsorsby:
I've just manually run the Symfony vulnerability checker which has found the dev-dependency propel has a potential mysql injection vulnerability. Unfortunately, it doesn't look like there is a fix for propel 1.7.

> Symfony
>
> Security Check Report
> =============================
>
> 1 package has known vulnerabilities.
>
> propel/propel1 (1.7.1)
> ----------------------
>
> * [CVE-NONE-0001][]: SQL injection possible with limit() on MySQL
>
> [CVE-NONE-0001]: https://github.com/propelorm/Propel/issues/1052
>
> Note that this checker can only detect vulnerabilities that are referenced in the security advisories 
> database.
>Execute this command regularly to check the newly discovered vulnerabilities.

pierredup:
Since this project doesn't ship with a `composer.lock` file, there wouldn't be much benefit in using the security checker to check the installed dependencies.
It is up to the applications using this package to ensure they check their `composer.lock` for any installed dependencies that are vulnerable.

The best we can do here is to require roave/security-advisories (as done in #904) to ensure we don't install and test against any vulnerable package.

pierredup:
Fixed in #904

PR #798, opened 2019-03-03
ADDED security check

liamsorsby:
This should help show security vulnerabilities before they become and issue. Addresses #797


liamsorsby:
@makasim Unfortunately, I tried to fix the tests by updating the version of Doctrine ORM. However, version 2.6.3 fixes the 'continue 2' regression in PHP 7.3 but they have also removed support for PHP 7.0 😞

makasim:
We could remove support of 7.0
No problem

bin/release
```diff
@@ -1,56 +0,0 @@
-#!/usr/bin/env bash
-
-# TODO: prepare changelog
```
makasim: Why did you remove the file?

liamsorsby: @makasim I just ran `git rm -rm . —cached && git add .`
It seems that the file matches one of the gitignore patterns
I can fix the gitignore though if it’s required.

liamsorsby:
I’ll amend the PR to fix the gitignore and remove support for PHP 7.0 as long as your happy with it as security support has already ended.

bin/release
liamsorsby: I've reverted the deletion of this file

pierredup:
@liamsorsby `sensiolabs/security-checker` is abandoned and replaced by https://github.com/fabpot/local-php-security-checker. Can you please update the PR to use the new binary?

pierredup:
Closing in favour of #904.

PR #801, opened 2019-04-01
 [Stripe][FIX] LogicException when card is declined

BadPixxel:
With Stripe, if "save_card" is requested but card is declined, CreateCustomer fails and both $model['card'] && $model['customer'] are unset. 
Fix => keep card unchanged if CreateCustomer fails and move forward

### DIFF
--- a/src/Payum/Stripe/Extension/CreateCustomerExtension.php
+++ b/src/Payum/Stripe/Extension/CreateCustomerExtension.php
@@ -80,14 +80,11 @@ class CreateCustomerExtension implements ExtensionInterface
 
         $gateway->execute(new CreateCustomer($customer));
 
-        $local['customer'] = $customer->toUnsafeArray();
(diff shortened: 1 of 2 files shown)


makasim:
@BadPixxel Thank for the contribution. Could you please add a test?

BadPixxel:
Let try! I'll update the existing test that set Status as failed as it's now unused

BadPixxel:
Look like tests are now OK.

BadPixxel:
Hi @makasim, Any reason not to merge this?

Great-Antique:
See #854

pierredup:
@BadPixxel If you are up to finishing this one, can you please rebase against master? That should fix the Twig deprecation notices and make the build green

BadPixxel:
@pierredup no problem I'll do that.

pierredup (approved):

PR #805, opened 2019-06-24
Added setAppInfo details to Stripe API requests

aliriaz-stripe:
- Set constants for Payum Stripe name, url and version
- Set `setAppInfo` before every Stripe API request

Please feel free to recommend edits/changes to this PR, happy to adjust as needed!

### DIFF
--- a/src/Payum/Stripe/Action/Api/CreateChargeAction.php
+++ b/src/Payum/Stripe/Action/Api/CreateChargeAction.php
@@ -8,6 +8,7 @@ use Payum\Core\Bridge\Spl\ArrayObject;
 use Payum\Core\Exception\LogicException;
 use Payum\Core\Exception\RequestNotSupportedException;
 use Payum\Core\GatewayAwareTrait;
+use Payum\Stripe\Constants;
 use Payum\Stripe\Keys;
 use Payum\Stripe\Request\Api\CreateCharge;
 use Stripe\Charge;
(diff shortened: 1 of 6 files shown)


src/Payum/Stripe/Constants.php
```diff
@@ -9,6 +9,12 @@ class Constants
 
     const STATUS_FAILED = 'failed';
 
+    const PAYUM_STRIPE_APP_NAME = "Payum Stripe";
+
+    const PAYUM_STRIPE_VERSION = "1.4.0";
```
pierredup, line 14: I think this rather needs to use [Composer\InstalledVersions::getVersion](
https://getcomposer.org/doc/07-runtime.md#knowing-the-version-of-package-x), so that we don't have to update this constant with each new release.

pierredup (changes requested):
@aliriaz-stripe Thanks for this contribution. Just a small comment.
Can you also rebase against master please.

src/Payum/Stripe/Action/Api/CreateChargeAction.php
```diff
@@ -65,6 +66,14 @@ public function execute($request)
         try {
             Stripe::setApiKey($this->keys->getSecretKey());
 
+            if (method_exists(Stripe::class, "setAppInfo")) {
```
pierredup, line 69: The next release of Payum will use version 7 of the Stripe SDK (#839), so this method should always exist then, and this check can be removed.

PR #817, opened 2019-09-06
Fix travis ci test by upgrading doctrine/orm and allow deprecations

alexander-schranz:
doctrine/orm 2.6 is needed for php 7.3 tests and currently there are some vendor deprecations which will fail the CI which are now ignored as it looks like the library is not very active maintained so I would ignore them at current state.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -28,7 +28,7 @@
 
         <ini name="soap.wsdl_cache_enabled" value="0" />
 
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@makasim this is making the CI tests run again.

pierredup:
Most of these changes are already in the master branch, so closing this. If there are any more improvements that can be merged, please re-open the PR

PR #818, opened 2019-09-11
Added helpful comments to get-it-started.md

RevisionTen:
I've seen multiple examples where developers point to the skd_config.ini full path instead of its directory (me included). This should fix that misconception.


pierredup:
Closing in favor of #887 (Rebasing against master while keeping the original commit)

PR #820, opened 2019-10-03
Allows to configure paypal rest sdk via config-array instead of ini file

florianorben:
Allows to set PayPal configuration by handling an array to the factory instead of relying on config_path to parse PayPal's skd_config.ini.


pierredup:
@florianorben The Travis failures are due to deprecations that are already fixed in Master. Can you please rebase from master to get the build green?

docs/paypal/rest/get-it-started.md
```diff
@@ -44,6 +44,30 @@ $payum = (new PayumBuilder())
     ->getPayum()
 ;
 ```
+alternatively set configuration via `config` see paypal's sdk_config.ini for available options
```
pierredup, line 47:
```suggestion

alternatively, set configuration via the `config` option. See PayPal's `sdk_config.ini` for available options.
```

src/Payum/Paypal/Rest/PaypalRestGatewayFactory.php
```diff
@@ -35,21 +35,30 @@ protected function populateConfig(ArrayObject $config)
...
+                    } elseif (PP_CONFIG_PATH !== $config['config_path']) {
+                        throw new InvalidArgumentException(sprintf('Given "config_path" is invalid. Should be equal to the defined "PP_CONFIG_PATH": %s.', PP_CONFIG_PATH));
+                    }
                 }
 
                 $credential = new OAuthTokenCredential($config['client_id'], $config['client_secret']);
-                return new ApiContext($credential);
+                $apiContext = new ApiContext($credential);
+
+                if (isset($config['config']) && is_array($config['config'])) {
```
pierredup, line 57: I think this will always evaluate to `true` since you are already setting the config option to an empty array in the default options. So maybe this check can just be removed, or maybe you can rather check that the config array is not empty?

src/Payum/Paypal/Rest/Tests/PaypalRestGatewayFactoryTest.php
```diff
@@ -72,6 +72,38 @@ public function shouldAllowCreateGateway()
...
+            'client_id' => 'cId',
+            'client_secret' => 'cSecret',
+            'config' => $givenConfig,
+        ]);
+
+        $apis = $this->readAttribute($gateway, 'apis');
+        $apiContext = null;
+        foreach ($apis as $api) {
+            if ($api instanceof \PayPal\Rest\ApiContext) {
+                $apiContext = $api;
```
pierredup, line 99: Once you have the required API, the loop can be broken

```suggestion
                $apiContext = $api;
                break;
```

src/Payum/Paypal/Rest/Tests/PaypalRestGatewayFactoryTest.php
```diff
@@ -72,6 +72,38 @@ public function shouldAllowCreateGateway()
...
+            }
+        }
+
+        $this->assertNotNull($apiContext);
+        $this->assertArraySubset($givenConfig, $apiContext->getConfig());
```
pierredup, line 104: `assertArraySubset` is deprecated in PHPUnit. The two arrays should be the same at this point AFAICS

```suggestion
        $this->assertSame($givenConfig, $apiContext->getConfig());
```

pierredup (changes requested):
Thanks for the changes @florianorben, just some minor comments here.

florianorben:
Hey @pierredup,
no problem, thanks for creating and maintaining this awesome lib :) 

All comments fixed!

src/Payum/Paypal/Rest/PaypalRestGatewayFactory.php
florianorben: Yes, you're right, it'll always evaluate to `true` - actually, after looking at the `\PayPal\Rest\ApiContext::setConfig` source code once again, I think it's fine to just remove that check.

src/Payum/Paypal/Rest/Tests/PaypalRestGatewayFactoryTest.php
florianorben: 👍  Good catch :)! Thanks

src/Payum/Paypal/Rest/Tests/PaypalRestGatewayFactoryTest.php
florianorben: Oh didn't realize it became deprecated , but your're right sadly.
I've replaced it with 

```
        $apiContextConfig = $apiContext->getConfig();
        foreach ($givenConfig as $k => $v) {
            $this->assertArrayHasKey($k, $apiContextConfig);
            $this->assertSame($v, $apiContextConfig[$k]);
        }
```

since the `ApiContext`'s config does have some default values (which are not required to be set via the factory), `config` and `$apiContext->getConfig()` are not necessarily the same.

pierredup:
> thanks for creating and maintaining this awesome lib :)

You should thank @makasim, I'm just reviewing PRs :)

pierredup (approved):
Looks good, thanks @florianorben

florianorben:
> You should thank @makasim, I'm just reviewing PRs :)

In that case - thanks to @makasim too :)!

makasim:
Thank you for contribution and review)

PR #836, opened 2020-02-11
symfony 5 compatibility

Tetragramat:
* Removed twig/twig 3.0 from composer because it is not actually compatible. Can be added back in another PR for twig 3 support.
* Changed symfony/* support to ^4.4|^5.0 same as in payum/payum-bundle.
* Changed doctrine/orm support to ^2.5 because only ^2.7 supports symfony 5.
* Updated phpunit/phpunit to ^7.5 because 5.7 had severe issues mocking classes with object typehint.
* Fixed error in Klarna Checkout in php 7.4.
* PHP version is in sync with minimal symfony version.


Tetragramat:
@makasim can you review/merge please?

makasim:
Thank you!

Issue #824, opened 2019-11-15
Attempted to call an undefined method named "__toArray" of class "Stripe\Charge". Did you mean to call "toArray"?


ghost:
A bug in CreateChargeAction.php (namespace Payum\Stripe\Action\Api) : 

Solved by : In line 70, replace "$charge->__toAttay" by "$charge->toArray"
![image](https://user-images.githubusercontent.com/48434889/68953133-8df5cc00-07c1-11ea-999c-78a548c35110.png)

yurguis:
@obrassart-karudev I'm using Symfony but I'm not getting this error you mention.

mrdev023:
> @obrassart-karudev I'm using Symfony but I'm not getting this error you mention.

I'm getting this error with Sylius

Prometee:
Since v7 of `stripe/stripe-php` the `__toArray` has been renamed to `toArray`, so `payum/payum` have to take care of this change and allow using both of those versions or drop support of older `stripe/stripe-php` (into the furtur v2.0 maybe)

pierredup:
This has been fixed in #827

PR #839, opened 2020-02-21
Upgrade to Stripe v7

joshbmarshall:
More complete PR to address #824 
Breaks backwards compatibility with v6 and previous versions of Stripe

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -73,7 +73,7 @@
         "paypal/rest-api-sdk-php" : "~1.0",
         "klarna/checkout": "^1.1 || ^2.0",
         "fp/klarna-invoice": "0.1.*",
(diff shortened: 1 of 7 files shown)


mikeymclellan:
+1 for merging this. In the meantime I have to checkout Josh's version if I want to run PHP7.4 as 7.4 requires Stripe v7

joshbmarshall:
@makasim is it possible to merge this, or do we need to release a major version as it may be backwards breaking for stripe v6 users ?

makasim:
Just set the desired minimum version of Stripe library in composer file.

joshbmarshall:
@makasim I changed the composer file, is that correct in require-dev or did you mean for it to be in the require section?

pierredup (approved):

PR #842, opened 2020-02-25
Add fields to set stripe variables for allow remember me and email

joshbmarshall:


src/Payum/Stripe/Resources/views/Action/obtain_checkout_token.html.twig
```diff
@@ -12,6 +12,8 @@
             data-description="{{ model.description|default("") }}"
             data-amount="{{ model.amount }}"
             data-currency="{{ model.currency|default("USD") }}"
+            data-allowRememberMe="{{ model.local.allowRememberMe|default("false") }}"
+            data-email="{{ model.local.email|default("") }}"
```
pierredup, line 16: Are these options still valid? I can't find anything about these options in the docs. It might only be relevant for an older version of Stripe

joshbmarshall: Yes the version of stripe checkout built into Payum is the legacy version, and the new version of stripe checkout is quite different.
It isn't a direct upgrade from what I can see, for simply taking payments. I just recently wrote a payment method using Stripe Elements to do a simple payment - https://github.com/joshbmarshall/payum_stripe

PR #843, opened 2020-02-27
Incorperated fix for redirect to capture.php from prepare.php from ht…

detain:
…tps://github.com/Payum/Payum/issues/600      otherwise when i loaded prepare.php it was redirecting me to like prepare.php/capture.php/capture.php/capture.php/capture.php/capture.php/etc..

### DIFF
--- a/docs/paypal/rest/get-it-started.md
+++ b/docs/paypal/rest/get-it-started.md
@@ -36,9 +36,9 @@ $payum = (new PayumBuilder())
     ->addDefaultStorages()
     ->addGateway('gatewayName', [
         'factory' => 'paypal_rest',
-        'client_id' => 'REPLACE IT',
-        'client_secret' => 'REPLACE IT',
-        'config_path' => 'REPLACE IT',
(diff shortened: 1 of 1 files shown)


src/Payum/Core/Bridge/PlainPhp/Security/TokenFactory.php
```diff
@@ -31,7 +31,17 @@ public function __construct(StorageInterface $tokenStorage, StorageRegistryInter
      */
     protected function generateUrl($path, array $parameters = [])
     {
-        $newPath = (new Path($this->baseUrl->getPath()))->withTrailingSlash().$path;
```
pierredup, line 34: I think this can be simplified to one line, then you don't need all the extra checks below:

```php
$newPath = UriModifier::replaceBasename($this->baseUrl, $path)->getPath();
```

@detain Do you mind making the change, and also adding a unit test?

pierredup (changes requested):

pierredup:
Closing in favour of #889

PR #851, opened 2020-04-07
Suppress warning on PHP ^7.2

egils:
Since PHP 7.2 ```count()``` yields a warning on invalid countable types [warning on invalid countable types](https://www.php.net/manual/en/function.count.php#refsect1-function.count-changelog). Added an ```is_array``` check before calling count on $params.


Tetragramat (approved):
@makasim this one is ok

pierredup (approved):

PR #853, opened 2020-04-16
doctrine/mongodb-odm 2.0 support

### LINKED ISSUE #852: doctrine/mongodb-odm 2.0 support
Hello! Using Payum with `doctrine/mongodb-odm: ^2.0` throws an exception
```
The mapping file vendor/payum/core/Payum/Core/Bridge/Doctrine/Resources/mapping/Token.mongodb.xml is invalid:                                         
  Line 5:0: Element '{http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping}field', attribute 'strategy': [facet 'enumeration'] The value 'NONE' is not an element of the set {'addToSet', 'pu  
  shAll', 'increment', 'set', 'setArray', 'atomicSet', 'atomicSetArray'}.                                                                                                                                
                                                                                                                                                                                                         
  Line 5:0: Element '{http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping}field', attribute 'strategy': 'NONE' is not a valid value of the atomic type '{http://doctrine-project.org/schemas  
  /odm/doctrine-mongo-mapping}storage-strategy'.                                                                                                                                                         
                                                                                                                                                                                                         
  Line 5:0: Element '{http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping}field', attribute 'id': The attribute 'id' is not allowed.
```

BoShurik:
Fixes #852

### DIFF
--- a/docs/storages.md
+++ b/docs/storages.md
@@ -166,7 +166,7 @@ $tokenStorage = new DoctrineStorage(
 ### Doctrine MongoODM.
 
 ```
-php composer.phar install "doctrine/mongodb": "1.0.*@dev" "doctrine/mongodb-odm": "1.0.*@dev"
+php composer.phar install "doctrine/mongodb-odm": "^2.0"
 ```
 
 ```php
@@ -218,7 +218,6 @@ use Doctrine\ODM\MongoDB\Mapping\Driver\AnnotationDriver;
 use Doctrine\ODM\MongoDB\Mapping\Driver\XmlDriver;
 use Doctrine\ODM\MongoDB\DocumentManager;
 use Doctrine\ODM\MongoDB\Configuration;
-use Doctrine\MongoDB\Connection;
 use Payum\Core\Bridge\Doctrine\Storage\DoctrineStorage;
 
 Type::addType('object', 'Payum\Core\Bridge\Doctrine\Types\ObjectType');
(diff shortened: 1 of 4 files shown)


davidmpaz:
Hello @BoShurik,

thanks for taking care of this one. Just a question, is this change related to a version upgrade only in mongo or enforced by other 3rd party dependency on your side?

In [docs](https://github.com/Payum/Payum/blob/1.6.0/docs/storages.md#doctrine-mongoodm) i see the suggested version to use is `1.0.*@dev`. Would you like to update documentation related to the version to use please?

thanks in advance and regards,
David

/cc @Tetragramat @bennukem:
I see this as some kind of breaking compatibility change that in my opinion should be included in a new release. Probably we should start tagging for that if it is not done yet, @makasim could you please help on how was it done in the past? Scheduling/Marking new changes for new releases.

BoShurik:
`doctrine/mongodb-odm-bundle` which supports `symfony/symfony: ^5.0` only supports `doctrine/mongodb-odm: ^2.0`

BTW, this code should work in `doctrine/mongodb-odm: ^1.3` as `id` attribute was deprecated [in `1.3.0`](https://github.com/doctrine/mongodb-odm/commit/2add05f0fdba5c1a4aca98c56bb727a6b46c47e7)

Tetragramat:
This mention should be changed too https://github.com/BoShurik/Payum/blob/doctrine-odm/src/Payum/Core/composer.json#L72

BoShurik:
@Tetragramat done

pierredup:
This has been superseded by #875. Thanks for proposing the change #BoShurik

PR #857, opened 2020-05-25
Upgrade of league URI

chadyred:
Upgrade of league URI due to conflict


Tetragramat:
Thank you for the PR.

Payum supports `php ^7.1.3` but `league/uri 6` works only with `php ^7.2`, so you'll need to keep support of `league/uri 5` too.
Of course fixing of tests that upgrade of `league/uri` broke is required before merge is possible.

pierredup:
The has been superseded by #874. Thanks for proposing the change @chadyred.

PR #860, opened 2020-07-28
fix(symfony doc): set correct command name for debugging

pierreboissinot:


davidmpaz:
Hi @pierreboissinot,

thanks for your time. I can confirm this change in docs. 
Still can not explain myself why travis is failing. I feel ashamed right now xD

Do you have an idea?

Best,
David

Tetragramat:
Travis fails due to release of `doctrine/persistence 2.0`. See https://github.com/Payum/Payum/issues/863

Current docs should not be wrong. Both `payum:gateway:debug` and `debug:payum:gateway` are valid command names. So I'm not sure if this PR is nessesary.

davidmpaz:
Means this that the change of command to use in docs are dependent of versions then? 
We could state that in the docs. Something like: starting on version x.x.x the command is name like `xyz`.

I have a Symfony 5 application and can confirm from error i got that:  *Command "payum:gateway:debug" is not defined.* and that `debug:payum:gateway` does exist.

Make sense?

Tetragramat:
It is supposed to work. If it does not work then it's bug. You should make an issue that command alias does not work.

`debug:payum:gateway` is defined as name of the command https://github.com/Payum/PayumBundle/blob/master/Command/DebugGatewayCommand.php#L21
`payum:gateway:debug` is defined as alias of the command https://github.com/Payum/PayumBundle/blob/master/Command/DebugGatewayCommand.php#L31

This PR changes command name in docs from "deprecated" `payum:gateway:debug` to preffered `debug:payum:gateway`. So I think it's ok to merge this.

@makasim This PR is ok. Travis fail is not caused by this PR.

Tetragramat (approved):

pierreboissinot:
@Tetragramat ok, I'l review this PR and check PayumBundle repo. Thanks.

pierredup:
Closing in favor of #886, while keeping the original commit (just rebased from master).
The name change to the preferred name is fine, while still keeping the alias for BC.

Issue #840, opened 2020-02-24
Twig 3.x compatibility


carlos-mg89:
Hi,

I'm using Symfony 5 along with Twig 3 and have found an issue while trying to set up Payum. Concretely, it seems that Payum is not compatible with Twig 3.x. If that is so, and bearing in mind that Twig 3.x has been there since 15 Nov 2019 (https://github.com/twigphp/Twig/releases/tag/v3.0.0), I believe it should be supported as well.

Any possibility this could be done in a short time? Or perhaps it involves a serious refactor?

I'm not aware of the changes between Twig 3 and the last supported Twig from Payum (v2.12.5).

Thanks in advance

carlos-mg89:
For anyone facing this issue (or an issue with Symfony 4-5 and Payum as well) just follow these steps:

1. Downgrade Twig 3 dependency to the latest Twig 2 version (at the moment, v2.12.5) with this: `composer require twig/twig:v2.12.5`
2. Try to install Payum with this command: `composer require "payum/payum-bundle" "payum/offline" "php-http/guzzle6-adapter"` although you can change the payum/offline for your payment gateway(s). In my case it was payum/stripe.
3. Last step will likely produce this error: `The child node "security" at path "payum" must be configured.` which can be fixed by creating the config/packages/payum.yaml config file as exposed here https://github.com/Payum/Payum/issues/787 and by creating the Payment and PaymentToken entities.
4. Rerun `composer require "payum/payum-bundle" "payum/offline" "php-http/guzzle6-adapter"` and this time there will be no errors

PR #861, opened 2020-09-09
Twig 3 and fix tests

StudioMaX:
Compatibility with Twig 3.

Closes https://github.com/Payum/Payum/issues/840, also closes https://github.com/Payum/Payum/issues/863


davidmpaz:
Hopefully, this will get merge soon.

src/Payum/Core/Registry/AbstractRegistry.php
```diff
@@ -52,10 +51,17 @@ public function getStorage($class)
         $class = is_object($class) ? get_class($class) : $class;
 
         // TODO: this is a quick fix. I have to find a better\clean solution.
-        if (class_exists($class) && interface_exists(Proxy::class)) {
-            $rc = new \ReflectionClass($class);
-            if ($rc->implementsInterface(Proxy::class)) {
-                $class = $rc->getParentClass()->getName();
+        if (class_exists($class)) {
+            if (interface_exists('Doctrine\Persistence\Proxy')) {
```
davidmpaz, line 55: Hi @StudioMaX allow me to ask, why not using here the approach: `Proxy::class` instead `'Doctrine\Persistence\Proxy'` as string? I think using `::class` is a more modern approach, don't you?

Thanks in advance for your help on this.

src/Payum/Core/Tests/Bridge/Doctrine/Storage/DoctrineStorageTest.php
```diff
@@ -134,18 +134,18 @@ public function shouldFindModelById()
     }
 
     /**
-     * @return \PHPUnit_Framework_MockObject_MockObject|\Doctrine\Common\Persistence\ObjectManager
+     * @return \PHPUnit_Framework_MockObject_MockObject|\Doctrine\Persistence\ObjectManager
      */
     protected function createObjectManagerMock()
     {
-        return $this->createMock('Doctrine\Common\Persistence\ObjectManager');
+        return $this->createMock('Doctrine\Persistence\ObjectManager');
```
davidmpaz, line 141: As a follow up to my previous comment. If you agree with me on using `::class` i would consider changing this also, it is already a change introduced and will help to keep consistency throughout the code.

davidmpaz (changes requested):
Hi @StudioMaX,

I made some review and added some comments about the usage of string to refer to class names instead using the static `class` attribute. would you mid please to provide some feedback. I believe afterward this PR can be merged.

Thanks in advance,
David

src/Payum/Core/Registry/AbstractRegistry.php
StudioMaX: As I remember this is because we need to check an existence of both `Doctrine\Persistence\Proxy` and `Doctrine\Common\Persistence\Proxy`, which means that we need to import one of them via an alias, like:
```php
use Doctrine\Persistence\Proxy;
use Doctrine\Common\Persistence\Proxy as ProxyOld;
...
if (interface_exists(Proxy::class)) {
   ...
elseif (interface_exists(ProxyOld::class)) {
   ...
```

In my opinion this is less readable than a plain old string.

davidmpaz: Fair enough. Thanks for clarifying it.

BoShurik:
Can you change `.travis.yml` like [this](https://github.com/Payum/Payum/blob/d241b25a48f6004b447ec12ca7ef3f02d7aa80ac/.travis.yml) to check tests with lower dependencies?

StudioMaX:
@BoShurik ok, I've added this lower dependency check.

Tetragramat (approved):

cariba:
Thank you @StudioMaX for this great fix!
Any idea when it will be available?
We're all in SF 5 and Twig 3 so we can't use payum without this being merged.

davidmpaz:
Hi @makasim,

this was approved already. If no more TODOs left, would you like to merge it please?

thanks in advance and regards,
David

Issue #863, opened 2020-10-01
Missing dependency on doctrine/persistence ^1.0


Tetragramat:
Due to missing dependency on `doctrine/persistence ^1.0` tests fail since release of `doctrine/persistence 2.0`.

> PHP Fatal error:  Interface 'Doctrine\Common\Persistence\Proxy' not found in /home/travis/build/Payum/Payum/src/Payum/Core/Tests/Registry/AbstractRegistryTest.php on line 341

Options:
1. Add dependency to `doctrine/persistence ^1.0`.
2. Add dependency to `doctrine/common ^2.4 || ^3.0` and change used interface to `Doctrine\Common\Proxy`

PR #864, opened 2020-10-07
Fix dependencies to pass tests

BoShurik:
Now all test are failing because of it
`PHP Fatal error:  Uncaught Error: Interface 'Doctrine\Common\Persistence\Proxy' not found`

Also add tests for testing lowest dependencies and fix related issues

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,9 @@ cache:
 matrix:
   include:
     - php: 7.1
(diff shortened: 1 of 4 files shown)


Tetragramat:
Related Issue https://github.com/Payum/Payum/issues/863

BoShurik:
@Tetragramat this issue also had been solved in #861

BoShurik:
Fixed in #861

PR #865, opened 2020-10-25
Update payum/core dependency

RealAestan:
Following this : https://github.com/Payum/PayumBundle/pull/519 update dependency to allow compatibility with newer versions of softwares. @Tetragramat

### DIFF
--- a/src/Payum/AuthorizeNet/Aim/composer.json
+++ b/src/Payum/AuthorizeNet/Aim/composer.json
@@ -20,11 +20,11 @@
         }
     ],
     "require": {
-        "payum/core": "^1.5",
+        "payum/core": "^1.6",
         "authorizenet/authorizenet": "^1.8"
     },
     "require-dev": {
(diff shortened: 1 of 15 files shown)


Tetragramat:
I don't see any need to change `payum/core` deps in each gateway. Could you explain why it is needed?

RealAestan:
@Tetragramat To be able to install payum/stripe for example, in a symfony 5 project with payum/payum-bundle. Or it will be in conflicts (see https://github.com/Payum/PayumBundle/pull/519). Or the only way would be to install all gateways with payum/payum.

Tetragramat:
Could you make bug report with your project's composer require and result of your composer update?

RealAestan:
I made it work with the following composer settings : 
        "payum/core": "dev-master as 1.6",
        "payum/payum-bundle": "dev-master",
        "payum/stripe": "^1.5",
I'm closing this PR.

PR #827, opened 2019-11-25
Fix #824 Issue (Attempted to call an undefined method named "__toArray" of class "Stripe\Charge". Did you mean to call "toArray"?)

mrdev023:

PR #866, opened 2020-10-27
Update CreateChargeAction.php

fabiolor:
changed $charge->__toArray() to $charge->toArray() to adapt to current Stripe implementation


pierredup:
This has been fixed in #827

PR #867, opened 2020-10-27
Update dev-master alias to 1.6-dev

jacquesbh:


dpfaffenbauer:
@jacquesbh can you please also update the aliases in the packages?

PR #871, opened 2020-12-31
Update PHPUnit tests

pierredup:
This is quite a large PR, but the main goal is to remove deprecations from PHPUnit and allow test runs on newer versions of PHPUnit. This also includes a few small improvements to the tests.

Due to the size of the PR, I think it's best to look at the individual commits. It shows all the different changes individually, and also makes it easier to revert a specific change if you don't agree with it.


makasim (approved):

makasim:
Looks good. Ready to merge ?

pierredup:
@makasim Yes this ons is ready to merge

makasim:
@pierredup thank you.

PR #873, opened 2020-12-31
Add support for PHP 8

pierredup:
Allow PHP 8 in composer, and run unit tests Travis against PHP 8.

~Edit: Seems there are some failures related to league/uri, will have a look at them and get the tests to pass~

Tests pass with no code changes necessary

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         }
     ],
     "require": {
-        "php": "^7.2",
+        "php": "^7.2 || ^8.0",
         "payum/iso4217": "~1.0",
         "php-http/message": "^1.0",
         "php-http/client-implementation": "^1.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -51,7 +51,7 @@
         "ext-pdo_sqlite": "*",
         "payum/omnipay-v3-bridge": "^1.0",
         "omnipay/dummy": "^3.0",
-        "omnipay/common": "^3.0",
+        "omnipay/common": "3.0.x-dev",
```
pierredup, line 54: There is no tagged version for `omnipay/common` yet that supports PHP 8, but support is added in the latest dev branch. So without this change, dependencies can't be reliably updated. Since this is just a dev dependency, it should be fine to keep it as an unbound constraint to allow the tests to run, which can then be updated as soon as there is a new tagged version

pierredup: This has been moved to Travis rather to only install dev dependencies in the CI

PR #879, opened 2021-01-05
Deprecate usage of payum/iso4217 and use alcohol/iso4217

pierredup:
This PR replaces the usage of the `payum/iso4217` fork with the original repo `alcohol/iso4217`, while still keeping the OOP functionality.

This means there is one less repo for the Payum team to maintain while getting an up to date version of ISO4217 currencies.

This also means the https://github.com/Payum/iso4217 repository can be deprecated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ matrix:
     - php: 7.4
     - php: 8.0
       before_install:
-        - composer req --dev league/uri-components:2.x-dev omnipay/common:3.0.x-dev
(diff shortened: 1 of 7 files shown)


awadood:
so will the payum/iso4217 be removed from composer.json payum/payum as the alcohol/4217 has been added to it?

pierredup:
@awadood I think that can only happen in a major version, as it would be a BC break. So for now, we just deprecate the usage of the package, and remove it in V2

PR #888, opened 2021-03-23
Update alcohol/iso4217 which supports PHP8

toooni:


pierredup:
Thanks for proposing this change @toooni, but this has already been added in #884.

toooni:
@pierredup But this PR is for the payum/core (https://github.com/Payum/Payum/blob/master/src/Payum/Core/composer.json)

pierredup:
Ah, good catch, totally missed that :) Thanks.

PR #915, opened 2021-07-15
Fix subtree split on GH Actions

pierredup:


pierredup:
@makasim Can you please add an org secret for `GITHUB_TOKEN` with write access to all repositories? This will help automate the subtree split so that we can automatically push the subtree to the correct repositories

pierredup:
@makasim Any chance you can either give me admin access to manage the secrets, or add the secrets to the repos so that I can finish this automation, and add a few other ones?

PR #921, opened 2021-10-15
Remove dev dependendies in PHP 8 workflow to use stable versions instead

pierredup:


pierredup:
This just fixes the installation of dependencies in GH actions for PHP 8.
Build failures are addressed in #920

PR #924, opened 2021-12-02
Add support for Doctrine DBAL 3

### LINKED ISSUE #922: [doctrine/dbal:^2] Unknown column type "json_array" requested
With `doctrine/dbal:^2.7` an error occured while using the command `doctrine:schema:update` :

```
Unknown column type "json_array" requested. Any Doctrine type that you use   
  has to be registered with \Doctrine\DBAL\Types\Type::addType(). You can get  
   a list of all the known types with \Doctrine\DBAL\Types\Type::getTypesMap(  
  ). If this error occurs during database introspection then you might have f  
  orgotten to register all database types for a Doctrine Type. Use AbstractPl  
  atform#registerDoctrineTypeMapping() or have your custom types implement Ty  
  pe#getMappedDatabaseTypes(). If the type name is empty you might have a pro  
  blem with the cache or forgot some mapping information.
```
[GitHub Action triggerring this error >>](https://github.com/Sylius/Sylius/runs/4192723890?check_suite_focus=true#step:9:16)

I could be wrong but there is no way to change a doctrine mapping field type outside of this package, right ?

So is it still a BC break according the comment found here : https://github.com/Payum/Payum/pull/920#discussion_r729777976 ?
 I can't see any migration path because those types `json` and `json_array` are exactly the same types from a database perpective.

pierredup:
As pointed out by @Prometee in https://github.com/Payum/Payum/issues/922#issuecomment-984156298, switching `json_array` to `json` shouldn't be a BC break because we don't store `null` values in the DB.

Fixes #922

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,7 @@
         "payum/omnipay-v3-bridge": "^1.0",
         "omnipay/dummy": "^3.0",
         "omnipay/common": "^3.0",
-        "doctrine/dbal": "^2",
+        "doctrine/dbal": "^2.7 || ^3.0",
         "doctrine/orm": "^2.5",
         "doctrine/persistence": "^1.3.3 || ^2.0",
         "phpunit/phpunit": "^7.5 || ^8.5",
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -54,7 +54,7 @@
         "payum/omnipay-v3-bridge": "^1.0",
         "omnipay/dummy": "^3.0",
         "omnipay/common": "^3.0",
-        "doctrine/dbal": "^2",
+        "doctrine/dbal": "^2.0 || ^3.0",
```
stloyd, line 57:
```suggestion
        "doctrine/dbal": "^2.7 || ^3.0",
```
Otherwise with low deps, it will fail.

Chris53897:
@pierredup thanks for the PR.
Is there a relase planned with these changes? I guess many of us don't want to use dev-master for stability reasons.

PR #926, opened 2022-02-04
feat: add return hint for setLogger() to void

Chris53897:
Hi. This will fix this error i got with symfony 6 (and Upgrading psr/log (2.0.0 => 3.0.0))

`PHP Fatal error:  Declaration of Payum\Core\Bridge\Psr\Log\LoggerExtension::setLogger(Psr\Log\LoggerInterface $logger) must be compatible with Psr\Log\LoggerAwareInterface::setLogger(Psr\Log\LoggerInterface $logger): void in /vendor/payum/core/Payum/Core/Bridge/Psr/Log/LoggerExtension.php on line 34
`

There are probably more, but these are the first that i get.


Chris53897:
@pierredup Is there a planned date for a 1.7.1 tag that include this change?
This would help for https://github.com/Payum/PayumBundle/pull/541

PR #929, opened 2022-03-15
feat: fix symfony6 deprecation messages for getParent()

Chris53897:
Solution with annotations.
Should be changed to Return values with next major version


pierredup:
Thank you @Chris53897

Issue #928, opened 2022-03-14
Payum v2 with php 8.1 as minimum requirements


johnkrovitch:
Hi,

I use Payum with Sylius quite a lot. I'm satisfied with general working, but I think there is a place for improvement in the code, especially in the gateway working and the actions system. It can benefit from having strong typing, interfaces and in the general process design. It can use all the new features provided in PHP 8. But this will require a new major version as some old code should be broken.

Are you planning to do this ? Are you interested in a contribution to start the movement ?

Have a nice day ! Thank you !

- [ ] Update the composer.json of Payum main library
- [ ] Update the composer.json of Payum Bundle
- [ ] Update the composer.json of AuthorizeNet/Aim]
- [ ] Update the composer.json of Be2Bill
- [ ] Update the composer.json of Core
- [ ] Update the composer.json of Klarna
- [ ] Update the composer.json of Offline
- [ ] Update the composer.json of Payex
- [ ] Update the composer.json of Paypal
- [ ] Update the composer.json of Skeleton
- [ ] Update the composer.json of Sofort
- [ ] Update the composer.json of Stripe
- [ ] Integrate the New Language Level Features
- [ ] Add strong typing

Chris53897:
@johnkrovitch I like this idea. Doing all this is a lot of work. I think the best way is to split the steps.
A first step could be to move to php >= 8. Modify all the composer.json Files so all Parts get an installable set in dev-section. That is not the case yet. Integrate the New Language Level Features.
Second Step add strong typing and other BC breaks like you described.

johnkrovitch:
Thank you for your response. I will add a check list for thing to do and link issue when created

Chris53897:
Just to be sure. I am a not a maintainer. Just a user who has interest in improving this repo.

johnkrovitch:
Same here. I use a lot with Sylius, and will be very happy if this repo is improved. I will wait for a maintainer feedback before starting anything.

pierredup:
I'm generally in favour of adding more strong types across the codebase, and this is something that I've been thinking of for a while. I also want to be conscious about an upgrade path, even if we release a new major version, I want to keep the BC breaks as minimal as possible, to ease the migration for users to the next major version.
I think that also means that we can release a last minor version in the 1.x range which includes all the deprecation notices (and can be released at the same time as the next major version).

I would also prefer to have PHP 8 as the minimum version instead of 8.1, just to make the package more accessible for a wider audience.

[...]

PR #932, opened 2022-04-26
feat: drop support of php < 8, update phpunit 5/8 => 9

Chris53897:
As discussed in #928 this is the first task needed for modernization (in my point of view).

- drop support of php < 8
- update phpunit 5/8 => 9 (with migration of phpunit.xml.dist)

There is of course a lot of space for improvments. But i do not want to do, to many thinks at once.

@pierredup @johnkrovitch
Maybe you can branch from `master` a new develop-branch `2.x` for modernization.
So i can take new branch as base for this PR. Advantage: `master` stays clean for other changes.

From that point we can improve that branch by ...

[ ] Add Typehints on internal API (avoid BC breaks)
[ ] Use new Language Features of PHP 8
[ ] Remove external deprecations (PhpUnit, ...)
[ ] Move Travis to github actions (Run tests on each repo separately)
[ ] Update Dependencies: Example: Possible BC Break: php-http/guzzle6-adapter => php-http/guzzle7-adapter
[ ] TODO

And after all that is done, new Interfaces and more complex code changes can be done.

So this single PR ist not getting bigger and bigger.

WDYT?


Chris53897:
Tests are failing because of deprecation warnings for PhpUnit 10.
GitHub Actions need to be adjusted for that.

pierredup:
@Chris53897 Thanks for working on this! I agree this PR was just getting too big. I have made some smaller changes on the master branch (updating version to 2.x, removing PHP 7 from composer, removing travis config file (since travis hasn't been used for a while) etc).

I have also reverted the commit that added all the types, so that we can focus only on the PHPUnit update in this PR.

Other changes can then be made in separate smaller PRs targeting the master branch. For things like adding type-hints, I also think it should be done incrementally and not all in one go (E.G take a few interfaces, add the types, and do the next few in another PR), just to make it easy to review and ensure we can revisit the types.
I want to try and avoid using the `mixed` type anywhere. If there is currently a method that accepts mixed, we should re-visit it to determine what type do we want, and if necessary, create the new interfaces or classes if we want to define a new type.

Chris53897:
@pierredup thanks for having a look. I am already working on an PR in the meantime for typehints. I will try to get tests working and will respond here after that is done.

pierredup (approved):

PR #947, opened 2022-09-30
PHP 8.2: Creation of dynamic property is deprecated

arokettu:
Fixes 8.2 warning: "Creation of dynamic property Payum\Core\Request\GetHttpRequest::$headers is deprecated"


pierredup:
Thanks @arokettu

Prometee:
@pierredup `payum/payum-bundle` is requiring this change on `1.*` versions since PHP 8.2 was launched.
Is there a reason why this PR has never been release into `1.*` version ?

PR #951, opened 2022-11-14
Added PayU India to supported gateway list.

knit-pay:
Added PayU India and PayUBiz to the supported gateway list.


pierredup:
Thank you @knit-pay

PR #956, opened 2023-03-06
Update get-it-started.md (forgot about namespace)

ibrahimsebbane:
to make the controller work in version 8, you have to put the namespace to the path of the controller


pierredup:
Thank you @ibrahimsebbane

Issue #838, opened 2020-02-13
Move Symfony integration from Payum/Core into Payum/PayumBundle


Tetragramat:
Payum/Core should not provide symfony integration when there is Payum/PayumBundle that does the same thing, but both do only part of the symfony integration.

Remove symfony integration from Payum/Core and move it into Payum/PayumBundle.

Tetragramat:
Thinking of making PR for this in near future but first I want to know if it would be accepted. @pierredup 

I would copy contents of https://github.com/Payum/Payum/tree/master/src/Payum/Core/Bridge/Symfony into root of https://github.com/Payum/PayumBundle fixed namespaces and references inside of PayumBundle. Next would be tests from https://github.com/Payum/Payum/tree/master/src/Payum/Core/Tests/Bridge/Symfony.
That would be all needed to do in PayumBundle PR.

What I'm not completely sure what to do in Payum/Core to keep it BC.
Maybe replacing contents of all classes with extend to PayumBundle's version?

Is there better solution than this?
```php
<?php
// src/Payum/Core/Bridge/Symfony/ReplyToSymfonyResponseConverter.php

namespace Payum\Core\Bridge\Symfony;

if (class_exists('Payum\Bundle\PayumBundle\ReplyToSymfonyResponseConverter')) {
  class ReplyToSymfonyResponseConverter extends Payum\Bundle\PayumBundle\ReplyToSymfonyResponseConverter {}
} else {
  class ReplyToSymfonyResponseConverter {
    // original content of the class
  }
}
```

pierredup:
@Tetragramat I think in payum/core, the classes can just be deprecated with a message specifying to use the classes in the bundle instead.

PR #985, opened 2024-02-08
feat: allow symfony 7 (WIP)

Chris53897:
There is probalby more work todo


Chris53897:
I am afraid i should have used 1.7.x as base branch

pierredup:
> I am afraid i should have used 1.7.x as base branch

The 1.7.x branch is only for bug fixes on the 1.7 release. New features should go into master.

Master is currently the 2.0.x development branch. There is no development branch for 1.8 since I'm not planning a 1.8 version release, so any new features should be added to master and will only be added to the version 2 release

pierredup:
Also, I think all the Symfony integration should be moved to Payum/PayumBundle and removed from Core (see #838), so it might be better to focus on the Symfony 7 support in the PayumBundle only

Chris53897:
Thanks for reply. I will try to find free time to work on this.

PR #989, opened 2024-03-08
feat: fix ci

Chris53897:
try to fix ci error on --prefer-lowest
One test is changed to fix this.

Additional the `"doctrine/annotations": "^1.14.3 || ^2.0.1", ` is now added.
I guess this dependency was made optional in some other dependency and was now missing.


Chris53897:
Errors for `rector` can be fixed in a follow up PR.

pierredup:
Thanks @Chris53897

PR #1004, opened 2024-09-26
Allow ^7.0 for league/uri and league/uri-components

mbenoukaiss:
According to the changelog none of the functions that got dropped in 7.0 version were used by payum/payum but some are seemingly missing from the changelog because the Path constructor being changed to private was not mentioned
https://github.com/thephpleague/uri/blob/master/CHANGELOG.md
https://github.com/thephpleague/uri-components/blob/master/CHANGELOG.md

If I were to make a similar pull request into 1.7.x would it be possible to accept it and make a new patch version or the fact it's a major version bump for a dependency would forbide that ?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,8 +44,8 @@
         "payum/iso4217": "~1.0",
         "php-http/message": "^1.0",
         "php-http/client-implementation": "^1.0",
-        "league/uri": "^6.4",
-        "league/uri-components": "^2.2",
+        "league/uri": "^6.4 || ^7.0",
+        "league/uri-components": "^2.2 || ^7.0",
         "twig/twig": "^1.34 || ^2.4 || ^3.0"
     },
     "require-dev": {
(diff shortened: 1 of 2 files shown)


pierredup:
Thanks for the patch @mbenoukaiss

> If I were to make a similar pull request into 1.7.x would it be possible to accept it and make a new patch version

Yes I think this can go into 1.7.x, you can just rebase your PR against 1.7.x. After merging and creating a new release, the changes should sync up to the master branch

mbenoukaiss:
So like this ?

Also for the PHPStan error I made a slight change to the way I check if the method exists by using an if block instead of a ternary seeing other similar checks in the project are made this way, but if it does not fix the warning, how should I proceed ? Use a phpstan ignore comment or use reflection to check if the constructor exists (if it even fixes the phpstanwarning) ? Or are there other ways ?

pierredup:
Thank you @mbenoukaiss

ownede:
Just ran into this issue myself. 
For people still having the issue: I wasn't able to require `payum/core` for `1.7.x` branch, as the dependencies weren't fixed there. A workaround is requiring `1.7.x` branch for `payum/payum`, which worked fine. 

@pierredup do you maybe have an idea when 1.7.6 could get released? 
It would be great to switch to a tagged version instead of dev branch.

PR #1022, opened 2025-07-22
fix league deprecations

dpfaffenbauer:


pierredup:
The tests are failing due to this change. I think these would need a check to only use the new methods if they exist, so that we can stay compatible with previous versions. For the master branch, we can drop support for older versions of league/uri

dpfaffenbauer:
thanks @pierredup ❤️

PR #1030, opened 2026-01-16
Refactor RequestTokenVerifier to address League/Uri deprecations

mcarbonell-paymefy:
This commit updates RequestTokenVerifier.php to ensure compatibility with newer versions of the league/uri library while maintaining backward compatibility.

Following the pattern established in previous core updates (commit 4efe88d), the following changes were made:
- Replaced deprecated static factory calls with the new new() method where available.
- Added method_exists checks for both HttpUri and Path classes to prevent breaking changes in environments using older versions of the library.
- Aligned imports and class aliases with existing core standards (using HttpUri as an alias for League\Uri\Http).

These changes eliminate deprecation notices in modern PHP environments and unify the URI handling logic across the project.

### DIFF
--- a/src/Payum/Core/Security/Util/RequestTokenVerifier.php
+++ b/src/Payum/Core/Security/Util/RequestTokenVerifier.php
@@ -3,7 +3,7 @@
 namespace Payum\Core\Security\Util;
 
 use League\Uri\Components\Path;
-use League\Uri\Http;
+use League\Uri\Http as HttpUri;
 
 class RequestTokenVerifier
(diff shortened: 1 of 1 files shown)


pierredup:
Thank you @mcarbonell-paymefy mcarbonell-paymefy

lruozzi9:
Hi @pierredup and @mcarbonell-paymefy, the 'new' method is different on Path from the previous behavior. The right new method to use is fromUri. This has been fixed in #1031.

PR #1031, opened 2026-01-19
Fix URI path creation in RequestTokenVerifier

lruozzi9:
This will fix a BC introduced in #1030.


pierredup:
Thank you @lruozzi9

